PR #1, opened 2015-09-09
Logstash to Filebeat implementation

ruflin:
This is the first implementation of Filebeat which copies over the code from logstash and migrates it to a Beat. The goal is to make Filebeat more stable through refactoring and adding tests.

This is still work in and also depends on the lumberjack implementation: https://github.com/elastic/libbeat/pull/56


LICENSE
```diff
@@ -0,0 +1,14 @@
+Copyright 2012–2015 Jordan Sissel and contributors.
```
urso: License corret?

ruflin: @urso I will adjust it and take the license of all other beats.

PR #5, opened 2015-09-11
Fixed calls to logp.Info

tsg:
In many places logp.Info was called with a module name, but only
logp.Debug accepts that. In some places I changed logp.Info with other
error level (err, warn or debug), in some places I just removed the
parameter.


ruflin:
Thanks for going through and also adapting the log levels.

PR #11, opened 2015-09-14
Deadtime Reimplementation as IgnoreAfter

ruflin:
Deadtime settings was removed by accident. It was now readded to make adding older log files working again. The flag is now called ignoreAfter. Some cleanup was done.


config/config.go
```diff
@@ -9,6 +9,12 @@ import (
 	"time"
 )
 
+// Defaults for config variables which are not set
+const (
+	RegistryFile = ".filebeat"
```
urso: DefaultRegistryFile ?

crawler/harvester.go
```diff
@@ -82,13 +82,17 @@ func (h *Harvester) handleReadlineError(lastTimeRead time.Time, err error) error
...
+		}
+
 		if info.Size() < h.Offset {
 			logp.Debug("harvester", "File truncated, seeking to beginning: %s", h.Path)
 			h.file.Seek(0, os.SEEK_SET)
 			h.Offset = 0
-		} else if age := time.Since(lastTimeRead); age > h.FileConfig.DeadtimeSpan {
-			// if lastTimeRead was more than dead time, this file is probably
-			// dead. Stop watching it.
+		} else if age := time.Since(lastTimeRead); h.FileConfig.IgnoreOlder != "" && age > h.FileConfig.IgnoreOlderDuration {
```
urso: consider a predicate function

crawler/prospector.go
```diff
@@ -78,6 +83,18 @@ func (p *Prospector) Prospect(resume *ProspectorResume, output chan *FileEvent)
 	// Seed last scan time
 	p.lastscan = time.Now()
 
+	if p.FileConfig.IgnoreOlder != "" {
+
```
urso: maybe some less newlines?

crawler/harvester.go
ruflin: That would mean setting it to about 290 years :-)

crawler/prospector.go
```diff
@@ -164,9 +182,9 @@ func (p *Prospector) scan(path string, output chan *FileEvent, resume *Prospecto
 			// Create a new prospector info with the stat info for comparison
 			newinfo = ProspectorInfo{Fileinfo: newFile.FileInfo, Harvester: make(chan int64, 1), Last_seen: p.iteration}
 
-			// Check for dead time, but only if the file modification time is before the last scan started
+			// Check for unmodified time, but only if the file modification time is before the last scan started
 			// This ensures we don't skip genuine creations with dead times less than 10s
-			if newFile.FileInfo.ModTime().Before(p.lastscan) && time.Since(newFile.FileInfo.ModTime()) > p.FileConfig.DeadtimeSpan {
+			if newFile.FileInfo.ModTime().Before(p.lastscan) && p.FileConfig.IgnoreOlder != "" && time.Since(newFile.FileInfo.ModTime()) > p.FileConfig.IgnoreOlderDuration {
```
urso: predicate function or bool variable with descripitve name to shorten line? Having OlderDuration to IntMax (hi possible) will simplify checks.

crawler/harvester.go
urso: I still interested in files from year 1725?

ruflin: I will go with the 1725

ruflin:
@urso updated accordingly

crawler/harvester.go
```diff
@@ -82,13 +82,17 @@ func (h *Harvester) handleReadlineError(lastTimeRead time.Time, err error) error
...
+		}
+
 		if info.Size() < h.Offset {
 			logp.Debug("harvester", "File truncated, seeking to beginning: %s", h.Path)
 			h.file.Seek(0, os.SEEK_SET)
 			h.Offset = 0
-		} else if age := time.Since(lastTimeRead); age > h.FileConfig.DeadtimeSpan {
-			// if lastTimeRead was more than dead time, this file is probably
-			// dead. Stop watching it.
+		} else if age := time.Since(lastTimeRead); age > h.FileConfig.IgnoreOlderDuration {
```
jordansissel: Should this use info.ModTime() instead of lastTimeRead?

crawler/prospector.go
```diff
@@ -78,6 +83,21 @@ func (p *Prospector) Prospect(resume *ProspectorResume, output chan *FileEvent)
...
 	p.lastscan = time.Now()
 
+	if p.FileConfig.IgnoreOlder != "" {
+
+		var err error
+		// Default ignore time time
+		p.FileConfig.IgnoreOlderDuration, err = time.ParseDuration(p.FileConfig.IgnoreOlder)
+
+		if err != nil {
+			logp.Warn("Failed to parse dead time duration '%s'. Error was: %s\n", p.FileConfig.IgnoreOlder, err)
```
jordansissel: What's the expected behavior if someone specifies an invalid time duration? A warning makes sense, but this doesn't seem to inform the user about what behavior is expected ("Failed to parse ... Using 24h as default") or maybe terminating due to a config failure?

ruflin: @jordansissel Good point. I still need to review all log messages to make sure the errors are properly communicated. In this case I actually suggest to shut down instead of using a default, as this could otherwise make a mess. I opened an issue so we don't forget about it: https://github.com/elastic/filebeat/issues/14

Issue #9, opened 2015-09-14
Implement Integration Tests


ruflin:
To better be able to test filebeat, integration tests must be added. This should be done similar to packetbeat with python and nosetest.

PR #13, opened 2015-09-14
Integration tests

ruflin:
Implementation for https://github.com/elastic/filebeat/issues/9


ruflin:
The current main issue of this pull request is that filebeat cannot be properly shut down because of all the channels. Some more refactoring is needed to decouple the components. This is happening here: https://github.com/elastic/filebeat/pull/15

tests/integration/test_crawler.py
```diff
@@ -0,0 +1,84 @@
...
+        with open(self.working_dir + "/output/filebeat") as f:
+            for i, l in enumerate(f, 1):
+                pass
+
+        # Check that output file has the same number of lines as the log file
+        assert iterations == i
+
+
+
+    def test_unifished_line(self):
```
tsg: typo in method name?

tests/integration/config/filebeat.yml.j2
```diff
@@ -0,0 +1,96 @@
...
+    {%- endif -%}]
+
+
+ # Uncomment the following if you want to ignore transactions created
+ # by the server on which the shipper is installed. This option is useful
+ # to remove duplicates if shippers are installed on multiple servers.
+ # ignore_outgoing: true
+
+{% if geoip_paths is not none %}
+ geoip:
```
tsg: Side comment: the geoip functionality should probably be separated from the libbeat/publisher, where I think it is now. It's Packetbeat specific in the current form.

So I think we won't need it in the template for FB.

tsg:
LGTM. Yeah, looks like the challenge with these tests will be how to reliably notice when filebeat has finished noticing the changes.

ruflin:
@tsg Agree. I'm very unhappy with the code that just waits for 10s to "assume" it is finished. Not sure if we could check the log and decide based on the output if it is done.

@monicasarbu The python code I used here is to 95% identical with packetbeat. Perhaps we can find a solution to "reuse" the code in a nicer way then just copy-paste.

ruflin:
@tsg Changes made as suggested.

Makefile
```diff
@@ -29,18 +30,28 @@ run: filebeat
...
+.PHONY: coverage
+coverage:
 	# gotestcover is needed to fetch coverage for multiple packages
 	go get github.com/pierrre/gotestcover
-	GOPATH=$(shell $(GODEP) path):$(GOPATH) $(GOPATH)/bin/gotestcover -coverprofile=profile.cov -covermode=count github.com/elastic/filebeat/...
-	mkdir -p cover
-	$(GODEP) go tool cover -html=profile.cov -o cover/coverage.html
+	mkdir -p coverage
+	GOPATH=$(shell $(GODEP) path):$(GOPATH) $(GOPATH)/bin/gotestcover -coverprofile=coverage/unit.cov -covermode=count github.com/elastic/filebeat/...
+	mkdir -p coverage
```
andrewkroh: Looks like this line is duplicated from above?

ruflin: You are right, I removed it.

monicasarbu:
Great, thanks!

PR #15, opened 2015-09-14
Refactor Crawler

ruflin:
The current implementation of Crawler, Prospector, Harvester passes around lots of state and is hard to change and test as it is tightly coupled. The goal of the refactoring is to take these blocks apart and split up in more independent blocks.

The changes made are mainly removing duplicated code and moving existing code around to simplify it.


crawler/crawler.go
```diff
@@ -0,0 +1,100 @@
...
+	cfg "github.com/elastic/filebeat/config"
+	"github.com/elastic/filebeat/input"
+	"github.com/elastic/libbeat/logp"
+	"os"
+)
+
+/*
+ The hierarchy for the crawler objects is explained as following
+
+ Crawler: Filebeat has one crawler. The crawler is the single point of control and stores the state. The state is written through the registrar
```
urso: somewhat long line. Even in github I have to scroll horizontally to read this.

crawler/crawler.go
```diff
@@ -0,0 +1,100 @@
...
+*/
+
+// Last reading state of the prospector
+type Crawler struct {
+	// List of all files which were crawled with the state
+	Files   map[string]*input.FileState
+	Persist chan *input.FileState
+}
+
+func (crawler *Crawler) Start(files []cfg.FileConfig, persist map[string]*input.FileState, eventChan chan *input.FileEvent) {
```
urso: same long line.

crawler/crawler.go
```diff
@@ -0,0 +1,100 @@
...
+ 		The harvester send their events to the spooler
+ 		The spooler sends the event to the publisher
+ 		The publisher writes the state down with the registrar
+*/
+
+// Last reading state of the prospector
+type Crawler struct {
+	// List of all files which were crawled with the state
+	Files   map[string]*input.FileState
+	Persist chan *input.FileState
```
urso: persist is used by prospector to update file states?

crawler/crawler.go
```diff
@@ -0,0 +1,100 @@
...
+			if pendingProspectorCnt == 0 {
+				break
+			}
+			continue
+		}
+		persist[*event.Source] = event
+		logp.Debug("prospector", "Registrar will re-save state for %s", *event.Source)
+	}
+
+	logp.Info("All prospectors initialised with %d states to persist", len(persist))
```
urso: so file state update's are only written after all prospectors finished?

crawler/prospector.go
```diff
@@ -83,49 +73,35 @@ func (p *Prospector) Prospect(resume *ProspectorResume, output chan *FileEvent)
...
-	} else {
-		logp.Debug("propsector", "Set ignoreOlderDuration to %s", cfg.DefaultIgnoreOlderDuration)
-		// Set it to default
-		p.FileConfig.IgnoreOlderDuration = cfg.DefaultIgnoreOlderDuration
-	}
-
 	// Now let's do one quick scan to pick up new files
 	for _, path := range p.FileConfig.Paths {
-		p.scan(path, output, resume)
+		p.scan(path, spoolChan)
```
urso: is a prospector passed around? Why not init callback-channel in Init?

crawler/prospector.go
```diff
@@ -6,64 +6,49 @@ import (
...
-	logp.Info("All prospectors initialised with %d states to persist", len(persist))
-
 }
 
-func (p *Prospector) Prospect(resume *ProspectorResume, output chan *FileEvent) {
-	p.prospectorinfo = make(map[string]ProspectorInfo)
+// Starts scanning through all the file paths and fetch the related files. Start a harvester for each file
+func (p *Prospector) Start(spoolChan chan *input.FileEvent) {
+
+	p.prospectorList = make(map[string]ProspectorFileStat)
```
urso: why is the make required here? is Prospect called multiple times for a prospector? Is Prospector Init incomplete?

crawler/prospector.go
```diff
@@ -83,49 +73,35 @@ func (p *Prospector) Prospect(resume *ProspectorResume, output chan *FileEvent)
...
+		p.scan(path, spoolChan)
 	}
 
 	// This signals we finished considering the previous state
-	event := &FileState{
+	event := &input.FileState{
 		Source: nil,
 	}
-	resume.Persist <- event
+	p.crawler.Persist <- event
```
urso: special 'close' event being send. If all Prospectors created by Crawler send this event, Crawler 'Start' exits.

beat/filebeat.go
```diff
@@ -88,14 +88,14 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
+	crawler := &Crawler{
 		Persist: make(chan *FileState),
 		// Load the previous log file locations now, for use in prospector
 		Files: make(map[string]*FileState),
 	}
 
-	registrar.LoadState(restart.Files)
-	restart.Scan(fb.FbConfig.Filebeat.Files, persist, fb.SpoolChan)
+	registrar.LoadState(crawler.Files)
+	crawler.Start(fb.FbConfig.Filebeat.Files, persist, fb.SpoolChan)
```
urso: is this blocking until Prospectors finish?

crawler/prospector.go
```diff
@@ -6,64 +6,49 @@ import (
...
 	}
-
-	logp.Info("All prospectors initialised with %d states to persist", len(persist))
-
 }
 
-func (p *Prospector) Prospect(resume *ProspectorResume, output chan *FileEvent) {
-	p.prospectorinfo = make(map[string]ProspectorInfo)
+// Starts scanning through all the file paths and fetch the related files. Start a harvester for each file
+func (p *Prospector) Start(spoolChan chan *input.FileEvent) {
```
urso: Name Start is irritating. Reading Start I would assume some go-routine being started. Instead Start is the inspectors main loop. Maybe call it 'Run'?

crawler/prospector.go
```diff
@@ -133,7 +109,8 @@ func (p *Prospector) Prospect(resume *ProspectorResume, output chan *FileEvent)
 	}
 }
 
-func (p *Prospector) scan(path string, output chan *FileEvent, resume *ProspectorResume) {
+// Scans the specific path which can be a glob (/**/**/*.log)
```
urso: What's scan doing with 'found' files? I see a channel for output + use of p.prospectorList update which is used as kinda output too.

crawler/prospector.go
```diff
@@ -143,8 +120,7 @@ func (p *Prospector) scan(path string, output chan *FileEvent, resume *Prospecto
 		return
 	}
 
-	// To keep the old inode/dev reference if we see a file has renamed, in case it was also renamed prior
-	missinginfo := make(map[string]os.FileInfo)
+	p.missingFiles = &map[string]os.FileInfo{}
```
urso: maps are already passed by reference. Why need a pointer to a map here?

crawler/prospector.go
```diff
@@ -6,64 +6,49 @@ import (
...
-}
-
 type Prospector struct {
 	FileConfig     cfg.FileConfig
-	prospectorinfo map[string]ProspectorInfo
+	prospectorList map[string]ProspectorFileStat
 	iteration      uint32
 	lastscan       time.Time
+	crawler        *Crawler
+	missingFiles   *map[string]os.FileInfo
```
urso: maps are already passed by reference. Why need a pointer to a map here?

crawler/prospector.go
```diff
@@ -169,143 +145,141 @@ func (p *Prospector) scan(path string, output chan *FileEvent, resume *Prospecto
...
-		p.prospectorinfo[file] = newinfo
+		p.prospectorList[file] = newInfo
 	} // for each file matched by the glob
 }
 
-func (p *Prospector) calculateResume(file string, fileinfo os.FileInfo, resume *ProspectorResume) (int64, bool) {
-	last_state, is_found := resume.Files[file]
+// Check if harvester for new file hast to be started
+// For a new file the following options exist:
+func (p *Prospector) checkNewFile(newinfo *ProspectorFileStat, file string, output chan *input.FileEvent) {
```
urso: that's funny. output is a channel of input.FileEvent :)

crawler/prospector.go
```diff
@@ -169,143 +145,141 @@ func (p *Prospector) scan(path string, output chan *FileEvent, resume *Prospecto
...
 
-	if previous := p.isFileRenamedResumelist(file, fileinfo, resume.Files); previous != "" {
-		// File has rotated between shutdown and startup
-		// We return last state downstream, with a modified event source with the new file name
-		// And return the offset - also force harvest in case the file is old and we're about to skip it
-		logp.Debug("prospector", "Detected rename of a previously harvested file: %s -> %s", previous, file)
-		last_state := resume.Files[previous]
-		last_state.Source = &file
-		resume.Persist <- last_state
-		return last_state.Offset, true
```
urso: more horizontal scrolling coming up :(

crawler/prospector_other.go
```diff
@@ -3,16 +3,16 @@
...
 
 import (
-	"github.com/elastic/filebeat/input"
 	"os"
 	"syscall"
 )
 
-func (p *Prospector) isFileRenamed(file string, info os.FileInfo, missingfiles map[string]os.FileInfo) string {
+// Checks if the given file was renamed. Returns the previous file on success, otherwise ""
+func (p *Prospector) isFileRenamed(file string, info os.FileInfo) string {
```
urso: Proposal:

// Check if the given file was renamed. If file is known but with different path,
// renamed will be set true and previous will be set to the previously known file path.
// Otherwise renamed will be false.
func (p *Prospector) isFileRenamed(file string, info os.FileInfo) (renamed bool, previous string)

crawler/prospector_other.go
```diff
@@ -3,16 +3,16 @@
...
 	// NOTE(driskell): What about using golang's func os.SameFile(fi1, fi2 FileInfo) bool instead?
```
urso: good question

crawler/prospector_other.go
```diff
@@ -23,20 +23,22 @@ func (p *Prospector) isFileRenamed(file string, info os.FileInfo, missingfiles m
...
 			return kf
 		}
 	}
+
+	// NOTE(ruflin): should instead an error be returned if not previous file?
 	return ""
 }
 
-func (p *Prospector) isFileRenamedResumelist(file string, info os.FileInfo, initial map[string]*input.FileState) string {
+func (c *Crawler) isFileRenamedResumelist(file string, info os.FileInfo) string {
```
urso: (optional) for Crawler method can also be named isFileRenamed.

crawler/prospector_other.go
```diff
@@ -23,20 +23,22 @@ func (p *Prospector) isFileRenamed(file string, info os.FileInfo, missingfiles m
...
 	// NOTE(driskell): What about using golang's func os.SameFile(fi1, fi2 FileInfo) bool instead?
```
urso: +1 for os independent os.SameFile

crawler/prospector_windows.go
```diff
@@ -13,7 +13,7 @@ func (p *Prospector) isFileRenamed(file string, info os.FileInfo, missingfiles m
 	return ""
 }
 
-func (p *Prospector) isFileRenamedResumelist(file string, info os.FileInfo, initial map[string]*input.FileState) string {
+func (c *Crawler) isFileRenamedResumelist(file string, info os.FileInfo, initial map[string]*input.FileState) string {
 	// Can we detect if a file was renamed on Windows?
 	// NOTE(driskell): What about using golang's func os.SameFile(fi1, fi2 FileInfo) bool?
 	return ""
```
urso: how about this: http://stackoverflow.com/questions/562701/best-way-to-determine-if-two-path-reference-to-same-file-in-windows

Same mechanism seems to be used by JDK.

go same file check for windows does should work as expected:
https://github.com/golang/go/blob/master/src/os/stat_windows.go#L29
https://github.com/golang/go/blob/f9ed2f75c43cb8745a1593ec3e4208c46419216a/src/os/types_windows.go#L55

etc/filebeat.yml
```diff
@@ -8,8 +8,14 @@ filebeat:
...
-      # Type of the files. Annotated in every documented
-      type: syslog
+      # Type of the files. Based on this the way the file is read is decided. The different types cannot be mixed.
+      #
+      # Possible options are:
+      # * log: Reads every line of the log file (default)
+      # * file: Sends the full file on change. This replaces fsRiver
+      # * meta: Reads new meta information on file change
+      # * stdin: Reads the standard in
+      type: log
```
urso: how much do we want to do for next beta? Beware of feature creep.

crawler/prospector_windows.go
```diff
@@ -13,7 +13,7 @@ func (p *Prospector) isFileRenamed(file string, info os.FileInfo, missingfiles m
...
 }
 
-func (p *Prospector) isFileRenamedResumelist(file string, info os.FileInfo, initial map[string]*input.FileState) string {
+func (c *Crawler) isFileRenamedResumelist(file string, info os.FileInfo, initial map[string]*input.FileState) string {
```
andrewkroh: The method signatures for the Windows methods (`isFileRenamedResumelist` and `isFileRenamed`) don't match up the signatures for Unix.

harvester/file.go
```diff
@@ -0,0 +1,4 @@
+package harvester
+
+// File harvester crawls full file and sends them to tika for "indexing".
```
tsg: What is tika?

harvester/log.go
```diff
@@ -218,3 +185,30 @@ func (h *Harvester) readline(reader *bufio.Reader, buffer *bytes.Buffer, eof_tim
 		}
 	}
 }
+
+// Handles eror during reading file. If EOF and nothing special, exit without errors
```
tsg: typo: s/eror/error

tsg:
Code reorg LGTM.

harvester/file.go
ruflin: https://tika.apache.org/ I will add a link to make it more obvious.

ruflin:
@urso @andrewkroh @tsg Thanks a lot for the review. Some of the comments I will directly implement into this pull request, some I will add as "TODO" in the comments to get done later. The reason is that otherwise this pull request will get bigger and bigger and prefer to do the refactoring step by step so errors can be better tracked in case something pops up later. I will comment for each "change request" what happens with it.

crawler/crawler.go
ruflin: shortened

crawler/crawler.go
ruflin: yes, better name is needed here. I have noted it in the source code.

crawler/crawler.go
ruflin: I broke it into two lines. I somehow don't like the identation that golang does for multi line arguments, it makes it always hard to read.

crawler/crawler.go
ruflin: To be honest I still need to analyse the registrar part in detail. This "docs" are taken over from logstash-forwarder.

crawler/prospector.go
ruflin: Good to know. Refactored.

crawler/prospector.go
ruflin: Renamed to Run.

crawler/prospector.go
ruflin: I moved it to Init() as this is where it belongs.

crawler/prospector.go
ruflin: Agree. Will do this in a next refactoring step. I marked it in the code.

crawler/prospector.go
ruflin: As far as I can see, all Prospectors send this event, but Crawler does not exit. Not sure I get why this should be the case?

crawler/prospector.go
ruflin: I extended the documentation of the method and I will renamed this "output" later so it is also more obvious what it is doing. The prospectorList is specific for every prospector and is non persistent.

crawler/prospector.go
ruflin: adjusted.

crawler/prospector.go
ruflin: Renaming/refactoring of further channels to make it more descriptive will happen at the next refactoring stage.

crawler/prospector.go
ruflin: adjusted

crawler/prospector_other.go
ruflin: Good one. Taken as suggested.

crawler/prospector_other.go
ruflin: I want to do some integration tests for this first before I replace the function to make sure the functionality is identical.

crawler/prospector_other.go
ruflin: Good point, done. This was not possible before as this were functions and not methods.

crawler/prospector_other.go
ruflin: First integration tests / unit tests, then I will change it.

crawler/prospector_windows.go
ruflin: Fixed.

crawler/prospector_windows.go
ruflin: Thanks for the reference. This looks good. I added a note to the source code.

etc/filebeat.yml
ruflin: Log is going to be the only one supported. But I want to have already in as I already structure the code for the other use cases which actually makes the code simpler.

harvester/log.go
ruflin: fixed

ruflin:
@urso @andrewkroh @tsg Changes applied

crawler/crawler.go
driskell: Originally, state would be loaded and harvesters started, then it would be discarded. This meant until all harvesters received new events and send to registrar, the state would be overwritten empty, then written with just one state, and so on, until all harvesters read data. This meant sometimes you could start forwarder, then restart too quickly and it would reingest log files. Just passing the loaded state back into registrar, however, would make another problem, where it would grow over time - though in hindsight I guess it's less a problem.

This was my first ever PR ever, and not the best code admittedly! The idea though was to wait for all prospectors to start and consider the loaded state, and pass back what should be kept, discarding anything that wasn't needed anymore (such as deleted files.) This final state was then passed to registrar.

crawler/prospector.go
urso: the crawler its start method is waiting for all prospectors it started to send this event to reduce the prospectorCnt (some kind of reference count). Only after all prospectors have send this event Crawler.start will return. With Crawler.start not run in its own go-routine, the spooler and publisher will only startup after the crawler.start returns.

Is startup order correct or would it make sense to start spooler and publisher more early?

crawler/crawler.go
urso: yeah, don't like default indentation either (in my editor parameters do align with code block :/

I started to do multiline like this (but I'm a ~80columns guy):

    func (crawler *Crawler) Start(
        files []cfg.FileConfig,
        persist map[string]*input.FileState,
        eventChan chan *input.FileEvent,
    ) {

Now you can shuffle/add/remove parameters at will.

crawler/crawler.go
ruflin: @driskell Thanks for the insights. When I start refactoring the registrar I will take this into account.

crawler/prospector.go
ruflin: @urso I will look into this. One is the startup order, the other part is the stop order. In the end what is important from my perspective that we have "at least once" sent for every log line.

PR #16, opened 2015-09-15
Return the error rather than logging it (and otherwise ignoring it)

tsg:
This should fix elastic/libbeat#75, should be merged after merging elastic/libbeat#77 and we need to update the Godeps as well.


ruflin:
Looks good to me. Will you update the dependencies directly in this pull request?

tsg:
@ruflin godeps updated.

PR #18, opened 2015-09-16
Integration tests

ruflin:
Add more integration test to check if filebeat works as expected.

### DIFF
--- a/tests/integration/test_crawler.py
+++ b/tests/integration/test_crawler.py
@@ -5,9 +5,10 @@ import time
 
 
 class Test(TestCase):
-    def test_crawler(self):
+    def test_fetched_lines(self):
+        # Checks if all lines are read from the log file
+
         self.render_config_template(
-            #path="/var/log/*"
             path=os.path.abspath(self.working_dir) + "/log/*"
         )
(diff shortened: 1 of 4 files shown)


tests/integration/test_crawler.py
```diff
@@ -3,11 +3,13 @@
 import os
 import time
 
-
+# Additional tests to be added:
+# * Check what happens when file renamed -> no recrawling should happen
```
andrewkroh: Is that not what `test_file_renaming` does?

tests/integration/test_crawler.py
```diff
@@ -83,3 +83,55 @@ def test_unfinished_line(self):
         assert iterations == i
 
 
+    def test_file_renaming(self):
+        # Makes sure that when a file is renamed, the content is not read again.
+        # Checks that if a line does not have a line ending, is is not read yet
```
andrewkroh: Copy/paste? I think this command can be removed.

monicasarbu:
Looks good to me

andrewkroh:
Aside from the two minor comments on comments, LGTM.

ruflin:
@andrewkroh I will clean up the comments.

Issue #14, opened 2015-09-14
Implement config failure in case of wrong ignoreOlder


ruflin:
See: https://github.com/elastic/filebeat/pull/11/files#r39432365

monicasarbu:
@ruflin Can this be closed?

ruflin:
This can be close with the update of the libbeat dependencies in https://github.com/elastic/filebeat/pull/35.

Issue #19, opened 2015-09-16
Make Scan Frequency Configurable


ruflin:
Currently the prospector scan is done every 10 seconds. This should be configurable especially also for testing purpose. The current 10 seconds are kept as default but a configuration variable will be added.

PR #20, opened 2015-09-16
Scan frequency Implementation

ruflin:
See https://github.com/elastic/filebeat/issues/19

### DIFF
--- a/crawler/prospector.go
+++ b/crawler/prospector.go
@@ -117,9 +117,8 @@ func (p *Prospector) Run(spoolChan chan *input.FileEvent) {
 
 		p.lastscan = newlastscan
 
-		// Defer next scan for a bit.
-		// TODO: Implement config variable or find algorithm when to rescan https://github.com/elastic/filebeat/issues/19
-		time.Sleep(10 * time.Second) // Make this tunable
+		// Defer next scan for the defined scanFrequency
+		time.Sleep(p.FileConfig.ScanFrequencyDuration)
 		logp.Debug("prospector", "Start next scan")
 
 		// Clear out files that disappeared and we've stopped harvesting
--- a/crawler/prospector.go
+++ b/crawler/prospector.go
@@ -28,7 +28,8 @@ type ProspectorFileStat struct {
 }
 
 // Init sets up default config for prospector
-func (p *Prospector) Init() {
+func (p *Prospector) Init() error {
+
 	if p.FileConfig.IgnoreOlder != "" {
 
 		var err error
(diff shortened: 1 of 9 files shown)


ruflin:
This change also closes https://github.com/elastic/filebeat/issues/14

crawler/prospector_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	prospector := Prospector{
+		FileConfig: fileConfig,
+	}
+
+	assert.NotNil(t, prospector)
+
+	prospector.Init()
+
+	// Default values expected
```
tsg: Is this comment right? Seems like the code checks non-default values.

tests/files/config.yml
```diff
@@ -13,6 +13,7 @@ filebeat:
         review: 1
         type: log
       ignoreolder: 24h
+      scanfrequency: 10s
```
tsg: I see most config file options use camelCase, but `ignoreolder` and `scanfrequency` do not. Any reason for this?

crawler/prospector.go
```diff
@@ -44,8 +46,27 @@ func (p *Prospector) Init() {
 		p.FileConfig.IgnoreOlderDuration = cfg.DefaultIgnoreOlderDuration
 	}
 
+	if p.FileConfig.ScanFrequency != "" {
+
+		var err error
+		// Default ignore time time
```
tsg: This comment is unclear to me.

ruflin: I removed the comment completely as I think the code is self explanatory. Same comment was also above, so it seems to come from copy-paste.

crawler/prospector_test.go
ruflin: Changed the comment. You are right of course.

tests/files/config.yml
ruflin: The reason is that this one is tested and the others are not :-) I would prefer to have scanFrequency, but when it is `scanFrequency` in the config file, it is not correctly mapped to go objects. Do you know any trick here?

tsg: yeah, you can use struct tags, like this:
<pre>
type FileConfig struct {
	Paths                 []string
	Fields                map[string]string
	Input                 string
	IgnoreOlder           string `yaml:"ignoreOlder"`
	IgnoreOlderDuration   time.Duration
	ScanFrequency         string `yaml:"scanFrequency"`
	ScanFrequencyDuration time.Duration
}
</pre>

ruflin: I opened an issue for it to make sure we don't forget about it: https://github.com/elastic/filebeat/issues/23

PR #21, opened 2015-09-17
Unit Testing Harvester Functions

ruflin:


harvester/log.go
```diff
@@ -133,24 +133,20 @@ func (h *Harvester) open() *os.File {
...
-	var newline_length int = 1
-	start_time := time.Now()
+
+	isPartial := true
+	startTime := time.Now()
 
 	for {
 		segment, err := reader.ReadBytes('\n')
 
+		fmt.Println(segment)
```
urso: some debugging left over?

ruflin: yep, will remove it.

harvester/log_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"math/rand"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+)
+
+func TestReadLine(t *testing.T) {
+
+	absPath, err := filepath.Abs("../tests/files/logs/")
```
urso: readline simply uses io.Reader. Why do we have to create a file in unit test here?

harvester/log_test.go
```diff
@@ -0,0 +1,108 @@
...
+	// All files starting with tmp are ignored
+	logFile := absPath + "/tmp" + strconv.Itoa(rand.Int()) + ".log"
+
+	assert.NotNil(t, absPath)
+	assert.Nil(t, err)
```
urso: if your test really really needs this file to be present consider t.Fatalf(t, ...) if err != nil. This will abort your test.

tests/integration/Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 PROCESSES ?= 4
-TIMEOUT ?= 60
+TIMEOUT ?= 90
```
urso: ouch

tests/integration/filebeat.py
```diff
@@ -241,3 +241,31 @@ def flatten_object(self, obj, dict_fields, prefix=""):
...
                 result[prefix + key] = value
         return result
+
+    def get_dot_filebeat(self):
+        # Returns content of the .filebeat file
+        dotFilebeat = self.working_dir + '/.filebeat'
+        assert os.path.isfile(dotFilebeat) == True
+
+        with open(dotFilebeat) as file:
+            data = json.load(file)
```
urso: return json.load(file)

tests/integration/filebeat.py
```diff
@@ -241,3 +241,31 @@ def flatten_object(self, obj, dict_fields, prefix=""):
...
+
+        return data
+
+    def get_filebeat_output(self):
+        # Returns content of the filebeat output file as json array
+        outputFile = self.working_dir + '/output/filebeat'
+        assert os.path.isfile(outputFile) == True
+
+        data = []
+        file = open(outputFile)
```
urso: with open(outputFile) as file:
    return [json.loads(line) for line in file]

tests/integration/test_crawler.py
```diff
@@ -135,3 +135,103 @@ def test_file_renaming(self):
...
+
+        testfile =  self.working_dir + "/log/test.log"
+        file = open(testfile, 'w')
+
+        iterations = 5
+        for n in range(0, iterations):
+            file.write("disappearing file")
+            file.write("\n")
+
+        file.close()
```
urso: I see test file generation to be a common theme for filebeat.

how about:

    def make_testfile(file, *content):
        with open(file, 'w') as file:
            for l in content:
                file.write(l)

    make_testfile("file1", "disappearing file\n")
    make_testfile("file2", "disappearing file\r\n")
    make_testfile("file3",
        "line1\n",
        "line2\r\n",
        "line3\r")

tests/integration/test_crawler.py
```diff
@@ -135,3 +135,103 @@ def test_file_renaming(self):
...
+
+        proc = self.start_filebeat()
```
urso: filebeat = self.start_filebeat()

urso: I'd like to have start_<beat> function to return a beat wrapper object around proc to do some more stuff like start, stop, restart, send some REST call if beat has REST API

tests/integration/Makefile
ruflin: Jep, I'm working on a solution that I can run "filebeat" once and it will stop on its won. This should then not be required anymore, at least not on this size.

tests/integration/test_crawler.py
```diff
@@ -135,3 +135,103 @@ def test_file_renaming(self):
...
+            file.write("new file")
+            file.write("\n")
+
+        file.close()
+
+        time.sleep(5)
+
+        proc.kill_and_wait()
+
+        data = self.get_dot_filebeat()
```
urso: data = fileabeat.dotfile()

tests/integration/test_crawler.py
```diff
@@ -135,3 +135,103 @@ def test_file_renaming(self):
...
+        )
+        os.mkdir(self.working_dir + "/log/")
+
+        testfile =  self.working_dir + "/log/test.log"
+        file = open(testfile, 'w')
+
+        iterations = 5
+        for n in range(0, iterations):
+            file.write("disappearing file")
+            file.write("\n")
```
urso: file.write(5 * "disappearing file\n")

harvester/log_test.go
ruflin: Agree, this is not necessarly required as it could also be mocked. Still it feels very comfortable to me to test it with "real" files as this could break on different platforms. Ok to leave it in?

harvester/log_test.go
ruflin: Added.

tests/integration/filebeat.py
ruflin: adjusted

tests/integration/filebeat.py
ruflin: fixed

tests/integration/test_crawler.py
ruflin: I will add something similar later when I add more tests.

tests/integration/test_crawler.py
ruflin: I replaced all the "proc" with "filebeat". I think the above would be part of the "beats-manager".

tests/integration/test_crawler.py
ruflin: In case we introduce a filebeat "object"

harvester/log_test.go
urso: sure, we can keep it in. but still wonder if use-case with files is caught by integration tests.

ruflin: Yes, but no specific to realine but the hole binary.

tests/integration/test_crawler.py
ruflin: That works :-) Feels more like ruby :-D

ruflin: BTW: I will "shorten/refactor" all the integration tests later. As the tests grow it gets more obvious what is used multiple times.

ruflin:
@urso I will wait for https://github.com/elastic/filebeat/pull/20 to be merged as scanfrequency could solve the problem of the failing travis tests.

PR #22, opened 2015-09-17
Run integration tests from `make test`

tsg:
Also added a `unit` target that runs only the unit tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ run: filebeat
 
 .PHONY: unit
 unit:
-	$(GODEP) go test -short ./...
+	$(GODEP) go test ./...
 
 .PHONY: test
 test: unit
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -26,10 +26,14 @@ clean:
 run: filebeat
 	./filebeat -c etc/filebeat.dev.yml -e -v -d "*"
 
-.PHONY: test
-test:
+.PHONY: unit
+unit:
 	$(GODEP) go test -short ./...
```
ruflin: should we also run the "long" tests here? I'm asking because I would except `make test` to run the full suite.

tsg: For the Beats it currently doesn't matter as they have no "long" tests. For libbeat, that would mean that you need Elasticsearch and Redis when running `make test`. I'm not sure about that part.

ruflin: Jep, which we have with the docker environment. As now everyone can run these tests without setup, they should also be run locally before pushing.

tsg:
@ruflin Removed the `-short` param.

PR #26, opened 2015-09-18
Config refactoring

ruflin:
The following config options were refactored and removed from command line options:
* spoolSize
* idleTimeout
* harvesterBufferSize
* tailOnRotate

### DIFF
--- a/beat/filebeat.go
+++ b/beat/filebeat.go
@@ -98,8 +98,14 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	registrar.LoadState(crawler.Files)
 	crawler.Start(fb.FbConfig.Filebeat.Files, persist, fb.SpoolChan)
 
-	// Start spooler: Harvesters dump events into the spooler.
+	// Init and Start spooler: Harvesters dump events into the spooler.
 	spooler := NewSpooler(fb)
+	err := spooler.Init()
+
+	if (err != nil) {
+		fmt.Print("Could not init spooler: %s", err)
+	}
+
 	fb.Spooler = spooler
 
 	go spooler.Start()
@@ -136,7 +142,6 @@ func (fb *Filebeat) Stop() {
 func emitOptions() {
 	logp.Info("\t--- options -------")
 	logp.Info("\tconfig-arg:          %s", configDirPath)
-	logp.Info("\tidle-timeout:        %v", cfg.CmdlineOptions.IdleTimeout)
 	logp.Info("\tharvester-buff-size: %d", cfg.CmdlineOptions.HarvesterBufferSize)
 	logp.Info("\t--- flags ---------")
 	logp.Info("\ttail (on-rotation):  %t", cfg.CmdlineOptions.TailOnRotate)
--- a/beat/filebeat.go
+++ b/beat/filebeat.go
@@ -103,7 +103,7 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	err := spooler.Init()
 
 	if err != nil {
-		fmt.Print("Could not init spooler: %s", err)
+		logp.Err("Could not init spooler: %v", err)
 	}
 
 	fb.Spooler = spooler
--- a/beat/filebeat.go
+++ b/beat/filebeat.go
@@ -102,7 +102,7 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	spooler := NewSpooler(fb)
 	err := spooler.Init()
 
-	if (err != nil) {
+	if err != nil {
 		fmt.Print("Could not init spooler: %s", err)
 	}
 
--- a/beat/filebeat.go
+++ b/beat/filebeat.go
@@ -37,6 +37,7 @@ type Filebeat struct {
 	SpoolChan     chan *FileEvent
 	publisherChan chan []*FileEvent
 	RegistrarChan chan []*FileEvent
+	Spooler       *Spooler
 }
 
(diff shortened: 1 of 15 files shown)


beat/filebeat.go
```diff
@@ -97,8 +97,17 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
...
 
-	// Start spooler: Harvesters dump events into the spooler.
-	go fb.startSpooler(cfg.CmdlineOptions)
+	// Init and Start spooler: Harvesters dump events into the spooler.
+	spooler := NewSpooler(fb)
+	err := spooler.Init()
+
+	if err != nil {
+		logp.Err("Could not init spooler: %v", err)
+	}
```
tsg: Is it safe here to continue in case of errors?

beat/spooler.go
```diff
@@ -3,31 +3,69 @@ package beat
...
 
-// startSpooler Starts up the spooler and starts listening on the spool channel from the harvester
-// Sends then bulk updates to the publisher channel
-func (fb *Filebeat) startSpooler(options *cfg.FilebeatConfig) {
+		var err error
+
+		config.IdleTimeoutDuration, err = time.ParseDuration(config.IdleTimeout)
+
+		if err != nil {
+			logp.Warn("Failed to parse idle timeout duration '%s'. Error was: %s\n", config.IdleTimeout, err)
```
tsg: \n is not needed at the end of the logp functions. Also, it's good practice to print errors using %v instead of %s.

beat/spooler.go
```diff
@@ -3,31 +3,69 @@ package beat
...
+	}
+}
+
+func (spooler *Spooler) Init() error {
+	config := &spooler.Filebeat.FbConfig.Filebeat
+
+	// Set default pool size is set to 0
+	if config.SpoolSize == 0 {
+		config.SpoolSize = cfg.DefaultSpoolSize
+	}
```
tsg: Probably not important here, but it's better to use pointers and compare with nil to know if an option is missing from the configuration file. This way, you don't have the issue of not being able to set SpoolSize to 0 (which is no problem here).

tsg:
Apart from the comments above, LGTM.

beat/filebeat.go
ruflin: No, I added `return err` and will update libbeat to make sure error is catched.

beat/spooler.go
ruflin: fixed

beat/spooler.go
ruflin: @tsg I tried to implement that because I think it makes a lot of sense to do it for all config variables. I'm not sure if my implementation was correct, but as soon as the struct is created by Unmarshal, all variables inside the struct (also the missing ones) are set to their default values. That means the pointer is not nil anymore and in the above case 0 is already set. Is there an other way to do the above mentioned?

ruflin:
@tsg I didn't merge in the newest libbeat yet as it breaks some other things (events). I think we should do this change in a second pull request.

beat/filebeat.go
```diff
@@ -97,8 +97,18 @@ func (fb *Filebeat) Run(b *beat.Beat) error {
 	registrar.LoadState(crawler.Files)
 	crawler.Start(fb.FbConfig.Filebeat.Files, persist, fb.SpoolChan)
 
-	// Start spooler: Harvesters dump events into the spooler.
-	go fb.startSpooler(cfg.CmdlineOptions)
+	// Init and Start spooler: Harvesters dump events into the spooler.
+	spooler := NewSpooler(fb)
+	err := spooler.Init()
```
urso: I would NewXxx to always return an initialised value

beat/spooler.go
```diff
@@ -3,31 +3,69 @@ package beat
...
+}
+
+func NewSpooler(filebeat *Filebeat) *Spooler {
+	return &Spooler{
+		Filebeat: filebeat,
+		running:  true,
+	}
+}
+
+func (spooler *Spooler) Init() error {
```
urso: What's the big difference between new and Init? Why not initialize NewSpooler correctly

beat/spooler.go
```diff
@@ -3,31 +3,69 @@ package beat
...
-var stopSpooler = false
+type Spooler struct {
+	Filebeat *Filebeat
+	running  bool
+}
+
+func NewSpooler(filebeat *Filebeat) *Spooler {
+	return &Spooler{
+		Filebeat: filebeat,
+		running:  true,
```
urso: running = true? Really?

ruflin: Yes, by default it is running. I assume your point is that this should be set in the actual "run" method.

beat/filebeat.go
ruflin: @urso I think I actually split this up for testing to have New only create the struct and the initialisation of config variables is in init.

beat/spooler.go
ruflin: See above. I split this up for testing. New only creates the struct, Init reads the config variables which has external dependencies.

urso: Ah, I see. This moves my problem to: why not pass the config to the spooler? Or rename Init to Config. Just allocating (using &spooler) and calling Init won't get me a correctly initialized state, right?

urso: yes, I would set running=true in some kind of start method

beat/spooler.go
urso: yes, I would set running=true in some kind of start method

PR #31, opened 2015-09-18
Adding a crosscompile make target.

andrewkroh:
Go 1.5 supports cross compiling without having to bootstrap the Go tool chain so I didn't need to use Docker at all.

The platforms on which filebeat does not current compile have been commented out.


Makefile
```diff
@@ -16,11 +17,7 @@ check:
 .PHONY: clean
 clean:
 	gofmt -w .
-	-rm filebeat
-	-rm filebeat.test
-	-rm .filebeat
-	-rm profile.cov
-	-rm -r coverage
+	-rm -rf filebeat filebeat.test .filebeat profile.cov coverage bin
```
ruflin: We must check if profile.cov must still be removed. It should be now also in the coverage folder.

ruflin:
@andrewkroh We should use this also make nightly binaries available somehow.

andrewkroh:
@ruflin  Regarding nightlies, I think we should get beats-packer building in travis / Jenkins and setup the job to publish to S3. Then integrate filebeat into beats-packer to get filebeat nightlies with OS specific packaging.

PR #32, opened 2015-09-21
Refactor and Simplify File comparison

ruflin:
Make it more cross platform compatible

### DIFF
--- a/input/file.go
+++ b/input/file.go
@@ -59,3 +59,21 @@ func (f *File) IsRegularFile() bool {
 	}
 	return true
 }
+
+
+// IsSameFile checks if the given File path corresponds with the FileInfo given
+func IsSameFile(path string, info os.FileInfo) bool {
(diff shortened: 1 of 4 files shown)


input/file.go
```diff
@@ -66,3 +59,20 @@ func (f *File) IsRegularFile() bool {
 	}
 	return true
 }
+
+// IsSameFile checks if the given File path corresponds with the FileInfo given
+func IsSameFile(path string, info os.FileInfo) bool {
+	fileInfo, err := os.Stat(path)
+
+	if err != nil {
+		logp.Info("Error during file comparison: %s with %s", path, info.Name())
```
tsg: I suggest including the error message as well. Also, error level?

tsg:
LGTM.

input/file.go
ruflin: I added it.

PR #33, opened 2015-09-21
Refactor files config to prospectors config

ruflin:
With moving also more options to the "files" config option which are actually prospector config options, this config option was renamed to make it more obvious and logic between code and config file.


tsg:
LGTM, but requires merging.

ruflin:
@tsg :-( Lets merge https://github.com/elastic/filebeat/pull/32 first.

PR #37, opened 2015-09-21
Refactor FileState to be universal

ruflin:


input/file_other.go
```diff
@@ -5,8 +5,15 @@ package input
 import (
 	"github.com/elastic/libbeat/logp"
 	"os"
+	"syscall"
 )
 
+// Returns Inode and Device
+func fileIds(info *os.FileInfo) (uint64, uint64) {
+	fstat := (*info).Sys().(*syscall.Stat_t)
+	return fstat.Ino, uint64(fstat.Dev)
```
andrewkroh: Consider casting `fstat.Ino` to `uint64` for the platforms on which `Ino` is actually a `uint32` (which I think is true for freebsd).

input/file_windows.go
```diff
@@ -0,0 +1,69 @@
...
+// The field can be found here: https://github.com/golang/go/blob/master/src/os/types_windows.go#L78
+func getFileInode(event *FileEvent) uint64 {
+	index, _, err := getInfo(event)
+	// This should return fileStat.idxhi as this value is unique in windows for a file per volume
+	return index
+}
+
+// getFileDevice returns the device id of the given file
+// The field can be found here: https://github.com/golang/go/blob/master/src/os/types_windows.go#L77
+func getFileDevice(event *FileEvent) uin64 {
```
andrewkroh: s/uin64/uint64/

ruflin: Yes it did :+1:

crawler/prospector_other.go
```diff
@@ -41,10 +41,14 @@ func (c *Crawler) isFileRenamed(file string, info os.FileInfo) string {
 	stat := info.Sys().(*syscall.Stat_t)
 
 	for kf, ki := range c.Files {
+
```
urso: no need for this extra newline

ruflin: @urso Looks like I like empty lines much more then you do :-D

urso: I like em too to separate 'information'/blocks of processing steps, but sometimes they feel like overkill.

input/file.go
```diff
@@ -76,3 +83,8 @@ func IsSameFile(path string, info os.FileInfo) bool {
 func (f1 *File) IsSameFile(f2 *File) bool {
 	return os.SameFile(f1.FileInfo, f2.FileInfo)
 }
+
+// Returns Inode and Device
+func fileIds(event *FileEvent) (uint64, uint64) {
```
urso: maybe make this one platform specific and allow for errors? For windows getInfo should be fileIDs then

ruflin: @urso the called "subfunctions" are platform specific. I made it two sub functions as I think i will be able to use these also in other places. But I'm also not very happy yet with the result (work in progress).

ruflin:
I will move to use reflection to get the values directly.

ruflin:
@tsg Yes that is a good switch too :-) I will squash the commits above as it got a little bit messy. The current version seems to work, but it is not tested on windows yet :-(

PR #39, opened 2015-09-21
Refactor registrar

ruflin:
Currently the state is in the crawler but should be part of the registrar. This will be a small refactoring.

### DIFF
--- a/crawler/registrar.go
+++ b/crawler/registrar.go
@@ -116,7 +116,7 @@ func (r *Registrar) fetchState(filePath string, fileInfo os.FileInfo) (int64, bo
 		return lastState.Offset, true
 	}
 
-	if previous := r.isFileRenamed(filePath, fileInfo); previous != "" {
+	if previous := r.getPreviousFile(filePath, fileInfo); previous != "" {
 		// File has rotated between shutdown and startup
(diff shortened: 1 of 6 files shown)


monicasarbu:
LGTM

crawler/registrar.go
```diff
@@ -0,0 +1,139 @@
...
+
+	r := &Registrar{
+		registryFile: registryFile,
+	}
+	r.Init()
+
+	return r
+}
+
+func (r *Registrar) Init() {
```
urso: why not Init(registryFile string) ?

Then New becomes:

    r := Registrar{}
    r.Init(registryFile)
    return &r

urso: Init is exported. => ensure Init on zero value has same result as using New?

monicasarbu: I agree with @urso.

ruflin: Also an option. I was thinking that init just does the default values, and NewRegistrar needs to be used to spcify a specific files. Probably we should have a general convention across the hole project for such things. Something like: X method is expected to do Y, Z is expected to do Q.

urso: +1

andrewkroh: Agree, a convention would be great.

PR #43, opened 2015-09-22
OS independent code for SafeFileRotate

tsg:
The two windows and !windows versions were similar with the windows version doing an extra move from .filebeat to .filebeat.old. This seems unnecessary and causes the rotation to fail on Win because
.filebeat doesn't exist (.filebeat.new does).

Don't merge yet, i'm still to test on windows.


tsg:
This doesn't work on windows :-(. Coming up with another solution.

driskell:
On Linux the rename overwrites. On Windows the target can't exist so it needs removing first - but atomicity is lost.

Maybe call `MoveFileEx` directly with `MOVEFILE_REPLACE_EXISTING`.
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365240(v=vs.85).aspx

driskell:
(Assuming that was the problem)

tsg:
@driskell yes, exactly. And right, MoveFileEx would solve the issue if we could call it from Go, which I couldn't find a solution for (that doesn't involve Cgo).

tsg:
I opened #45 with a less elegant and not atomic solution, but that works in my tests.

driskell:
Ah yes, well that explains why I didn't do it in courier!

PR #50, opened 2015-09-22
Improve documentation of default config file

ruflin:
…e config option


etc/filebeat.yml
```diff
@@ -2,59 +2,74 @@
...
-  cpuProfileFile:
-  registryfile: .filebeat
+  #spoolSize: 1024
+
+  # Defines how often the spooler is flushed. After idleTimeout the spooler is
+  # Flush even though spoolSize is not reached.
+  #idleTimeout: 5s
+
+  # Name of the registry file. Per default it is put in the same directory as
+  # the executable binary. In case this is changed after running filebeat,
```
tsg: I was wondering about this. Is it really in the same directory as the binary, or in the current working directory? The latter would be not ideal because the init script probably sets it to `/`.

etc/filebeat.yml
```diff
@@ -2,59 +2,74 @@
...
-  # List of prospectors to fetch data
+  # List of prospectors to fetch data.
   prospectors:
+    # Each - is a prospector. Below are the prospector specific configurations
     -
-      # Paths that should be crawled and fetched
+      # Paths that should be crawled and fetched. Glob based paths.
+      # For each file found under this path, a harvester is started.
       paths:
         - /var/log/*.log
```
tsg: `/var/log` exists on all platforms we currently support, except Windows, right? Not sure if we can have a "working" default for windows.

ruflin: Jep. I would probably suggest to leave it in as "example". On Windows you will get an error I assume.

tsg: We can provide a different example for Windows. For Packetbeat, for example, we change the default interface name depending on the OS at packaging time. But in this case there doesn't seem to be any good default. Maybe `c:\programdata\filebeat\logs\*` in which case it would index it's own logs (dangerous?). Or `c:\programdata\elasticsearch\logs\*` as an example.

ruflin: I suggest we put the second one with # in front into the config. So it is delete friendly for linux and easy to enable for windows.

tsg: +1.

etc/filebeat.yml
tsg: I did a quick test, looks like it's the current working directory, not the location of the binary. So the comment is not quite precise. We should consider adjusting this default, but in a different PR.

PR #51, opened 2015-09-22
Appveyor configuration

tsg:
Runs unit and integration tests on the Appveyor service.

The approach taken installs Go and python and then run the equivalent commands that the Makefile would run for building and testing. This means a bit of code duplication but using make on Windows is hard enough and maintaining a cross-platform Makefile would be a PITA.

Integration tests are currently failing due a genuine issue.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,51 @@
+# version format
+version: "{build}"
+
+# Operating system (build VM template)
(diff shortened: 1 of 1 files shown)


tsg:
Here is an example test run: https://ci.appveyor.com/project/tsg/filebeat/build/12

PR #52, opened 2015-09-22
Single cross-compile on travis-ci.

andrewkroh:
As suggested by @ruflin, this configures travis-ci to cross-compile filebeat only once (rather than running the cross-compile on each travis-ci OS).


ruflin:
@andrewkroh cool

PR #53, opened 2015-09-22
Fix reflection hack

tsg:
The ValueOf should receive the interface, not a pointer to the
interface. Then Elem() obtains a pointer to the struct, which is
just what's needed by FieldByName().

### DIFF
--- a/input/file_windows.go
+++ b/input/file_windows.go
@@ -1,9 +1,10 @@
 package input
 
 import (
-	"github.com/elastic/libbeat/logp"
 	"os"
 	"reflect"
+
+	"github.com/elastic/libbeat/logp"
 )
 
 type FileStateOS struct {
@@ -17,7 +18,7 @@ func GetOSFileState(info *os.FileInfo) *FileStateOS {
 
 	// Gathering fileStat (which is fileInfo) through reflection as otherwise not accessible
 	// See https://github.com/golang/go/blob/90c668d1afcb9a17ab9810bce9578eebade4db56/src/os/stat_windows.go#L33
-	fileStat := reflect.ValueOf(info).Elem()
+	fileStat := reflect.ValueOf(*info).Elem()
 
 	// Get the three fields required to uniquely identify file und windows
 	// More details can be found here: https://msdn.microsoft.com/en-us/library/aa363788(v=vs.85).aspx


driskell:
Bear in mind that the actual values aren't loaded by Golang unless you used Stat on an open file handle. Or at least not in previous versions.  

To get fileinfo on closed file (os.Stat) it doesn't get file info by handle this no IDs. For open file it uses the get info by handle. So fileinfo from os.Stat won't have any ID populated. May confuse a caller? Calling os.SameFile before reflection will ensure they are populated

https://golang.org/src/os/types_windows.go#L101

ruflin:
@driskell Thanks for pointing this out. So what you recommend is to call os.sameFile and comparing the file with itself to populate the information?

We should probably also add a small windows test to check this functionality to make sure we don't break it in the future.

driskell:
Yep if it's already populated it doesn't do anything but that seems the only way. Kind of makes the back hackier and I do wonder if your original method of just calling file info by handle directly was more long term stable and cleaner.

driskell:
I guess downside to calling it yourself though is you double the calls if you end up using SameFile somewhere else with the same fileinfo.

ruflin:
@driskell I added some unit tests here to run on Windows. Would you expect these to fail? https://github.com/elastic/filebeat/pull/56/files#diff-ff72076f4156619ba5ae2ea7ae72b36cR10

driskell:
That should pass I think. Does it?

ruflin:
Yep, it does: https://ci.appveyor.com/project/tsg/filebeat-dhcud/build/4

If I close the file and then call Stat it should fail I assume. What I'm not sure is which are the cases GetState() is called and the file is not open?

driskell:
Yes if you use os.Stat() it should fail.

Context is key, yes, so if the file info passed in is always from an open file, no problem. It's entirely up to you but I would say it could be confusing to a future developer who might try use it with file info from os.Stat() for whatever reason.

ruflin:
@driskell Spot on. I did some tests here and it is exactly as you said: 
* https://github.com/elastic/filebeat/pull/56/files#diff-ff72076f4156619ba5ae2ea7ae72b36cR29
* https://ci.appveyor.com/project/elastic-beats/filebeat/build/11

Looks like I will add the SameFile part. Thanks a lot for spotting this.

PR #64, opened 2015-09-24
Base vagrant file with Windows Environment

ruflin:


Vagrantfile
```diff
@@ -0,0 +1,26 @@
...
+  # Communicator for windows boxes
+  config.vm.communicator = "winrm"
+  #config.winrm.username = "admin"
+  #config.winrm.password = "beats"
+    
+  # Port forward WinRM and RDP
+  config.vm.network :forwarded_port, guest: 3389, host: 3389
+  config.vm.network :forwarded_port, guest: 5985, host: 5985, id: "winrm", auto_correct: true
+
+  # Load current folder as C:\dev directory
```
andrewkroh: I haven't used Vagrant + windows guest. It looks strange to have a unix path here. Does it convert this to a Windows path on the C drive? Just curious.

ruflin: Yes. I was thinking the same and it actually works

Vagrantfile
```diff
@@ -0,0 +1,26 @@
...
+# vi: set ft=ruby :
+
+### Documentation
+# This box is used as windows development and testing environment for filebeat
+# Two users exist: Administartor and Vagrant. Both have the password: vagrant
+
+Vagrant.configure(2) do |config|
+
+  # Windows Server 2012 R2
+  config.vm.box = "http://files.ruflin.com/vagrant/beats.box"
```
andrewkroh: Once you get the box finalized I think we should drop this in S3. What do you think?

ruflin: Jep, or on Atlas.

Vagrantfile
```diff
@@ -0,0 +1,26 @@
...
+  # Windows Server 2012 R2
+  config.vm.box = "http://files.ruflin.com/vagrant/beats-20150925.box"
+
+  # Communicator for windows boxes
+  config.vm.communicator = "winrm"
+  #config.winrm.username = "admin"
+  #config.winrm.password = "beats"
+
+  # Port forward WinRM and RDP
+  config.vm.network :forwarded_port, guest: 22, host: 22
```
andrewkroh: If this isn't used then remove. It it is then add auto_correct.

ruflin: I just remove the line and also the commented out lines above.

PR #65, opened 2015-09-24
Install config target and default template

tsg:
This adds the `install_cfg` Makefile target which is required by the beats-packer. Besides the configuration file, this is supposed to also install the default ES mapping template. I added a very basic template file, a more complete template should be worked out separately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ filebeat: $(GOFILES)
 	go get github.com/tools/godep
 	$(GODEP) go build
 
+
 .PHONY: check
 check:
 	# This should be modified so it throws an error on the build system in case the output is not empty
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -64,3 +65,14 @@ crosscompile: $(GOFILES)
 	go get github.com/tools/godep
 	mkdir -p bin
 	source crosscompile.bash; OUT='bin' go-build-all
+
+# This is called by the beats-packer to obtain the configuration file and
+# default template
+.PHONY: install_cfg
```
ruflin: I would suggest to call it install-cfg as with the full-coverage below.

tsg:
I posted #66 to follow up on the template discussion.

PR #67, opened 2015-09-28
Spooler Cleanup

ruflin:
Small spooler cleanup

### DIFF
--- a/beat/spooler.go
+++ b/beat/spooler.go
@@ -8,15 +8,24 @@ import (
 )
 
 type Spooler struct {
-	Filebeat *Filebeat
-	running  bool
+	Filebeat      *Filebeat
+	running       bool
+	nextFlushTime time.Time
+	spool         []*input.FileEvent
 }
 
 func NewSpooler(filebeat *Filebeat) *Spooler {
-	return &Spooler{
+	spooler := &Spooler{
 		Filebeat: filebeat,
 		running:  false,
 	}
+
+	config := &spooler.Filebeat.FbConfig.Filebeat
+
+	// Set the next flush time
+	spooler.nextFlushTime = time.Now().Add(config.IdleTimeoutDuration)
+
+	return spooler
 }
 
 func (spooler *Spooler) Config() error {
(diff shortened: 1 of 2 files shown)


beat/spooler.go
```diff
@@ -46,68 +57,64 @@ func (spooler *Spooler) Config() error {
...
-			break
+			}
 		}
 	}
 }
 
+// Stop stops the spooler. Flushes events before stopping
 func (s *Spooler) Stop() {
 	s.running = false
+	s.flush()
```
urso: possible race in shutdown? Worker thread + shutdown potentially calling flush at same time.

beat/spooler.go
```diff
@@ -46,68 +57,64 @@ func (spooler *Spooler) Config() error {
 	return nil
 }
 
+// Start starts the spooler
+// It heartbeats periodically. If the last flush was longer than
+// 'IdleTimeoutDuration' time ago, then we'll force a flush to prevent us from
+// holding on to spooled events for too long.
 func (s *Spooler) Start() {
```
urso: maybe rename 'Start' to 'Run' ?

beat/spooler.go
urso: Either have spooler worker to do final flush on shutdown or wait until spooler did quit (e.g. using WaitGroup). Consider both: sync.WaitGroup for shutdown and worker routine to do final flush.

beat/spooler.go
```diff
@@ -8,15 +8,26 @@ import (
 )
 
 type Spooler struct {
-	Filebeat *Filebeat
-	running  bool
+	Filebeat      *Filebeat
+	running       bool
+	nextFlushTime time.Time
+	counter       int
+	spool         []*input.FileEvent
```
urso: why not have nextFlushTime, counter and spool local to worker loop?

beat/spooler.go
```diff
@@ -46,68 +57,64 @@ func (spooler *Spooler) Config() error {
...
+	// Sets up ticket channel
 	ticker := time.NewTicker(config.IdleTimeoutDuration / 2)
 
-	// slice for spooling into
-	// TODO(sissel): use container.Ring?
-	spool := make([]*input.FileEvent, config.SpoolSize)
-
-	// Current write position in the spool
-	var spool_i int = 0
+	s.spool = make([]*input.FileEvent, config.SpoolSize)
```
urso: Consider more idiomatic slice-only implementation.

    s.spool = make([]*input.FileEvent, 0, config.SpoolSize)

So flush becomes:

    if len(s.spool) > 0 {
        // copy buffer
        tmp := make([]*FileInput, len(s.spool))
        copy(tmp, s.spool) // copy buffer

        // clear buffer
        s.spool = s.spool[:0]

         // send
         s.Filebeat.publisherChan <- tmp
    }

Adding a value to the pool becomes:

    s.spool = append(s.spool, event)

No need for s.counter

beat/spooler.go
ruflin: Renamed

beat/spooler.go
ruflin: I put it into the Struct because like this I can use it in different functions like flush without having to pass lots of variables every time.

beat/spooler.go
ruflin: Nice one. Done.

beat/spooler.go
ruflin: What about adding `defer s.flush()` to the Run function and remove it from Stop()? This should also prevent the race conditions.

ruflin:
@urso I applied the changes and squashed the commits into 1.

PR #56, opened 2015-09-23
Windows fixes

ruflin:
NOTE: Do not merge this pull request as this is a playground for testing out some windows file changes.

Some more windows fixes and tests for OSFileState

PR #74, opened 2015-10-06
Implements ReadOpen to make read only files available on windows

ruflin:
The problem on windows is that files which are already opened for reading cannot be read by and other process by default. This implements special flags to make the file available for reading on Windows.

In the future this should be extended also for files which have to be written to and moved into libbeat. See also: https://github.com/elastic/filebeat/pull/56

### DIFF
--- a/input/file_windows.go
+++ b/input/file_windows.go
@@ -1,8 +1,10 @@
 package input
 
 import (
+	"fmt"
 	"os"
 	"reflect"
+	"syscall"
 
 	"github.com/elastic/libbeat/logp"
 )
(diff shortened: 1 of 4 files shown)


input/file_windows.go
```diff
@@ -67,3 +69,42 @@ func SafeFileRotate(path, tempfile string) error {
 	}
 	return nil
 }
+
+// ReadOpen to opens a file for reading only
```
urso: s/to//

input/file_windows.go
```diff
@@ -67,3 +69,42 @@ func SafeFileRotate(path, tempfile string) error {
...
+// As Windows blocks deleting a file when its open, some special params are pass here.
```
urso: s/pass/passed/

tsg:
Nice! \o/

DanielLaberge:
I tried using the latest nightly build but logbeat still locks log files. Am I missing something?

https://beats-nightlies.s3.amazonaws.com/filebeat/filebeat-nightly.latest-windows.zip

ruflin:
@DanielLaberge Which Windows version are you using and what file system type?

DanielLaberge:
@ruflin Windows 8.1 Pro x64 with an NTFS file system.

DanielLaberge:
#156 Fixed the issue for me. Thanks!

PR #75, opened 2015-10-06
Cleanup open Todo's from Refactoring

ruflin:


beat/spooler.go
```diff
@@ -13,6 +13,7 @@ type Spooler struct {
 	running       bool
 	nextFlushTime time.Time
 	spool         []*input.FileEvent
+	SpoolChan     chan *input.FileEvent
```
urso: maybe rename to 'ch' or 'channel' or 'queue'? spooler.SpoolChan sounds kinda redundant.

beat/spooler.go
```diff
@@ -96,6 +95,10 @@ func (s *Spooler) Run() {
 			}
 		}
 	}
+
+	// Flush again before exiting spooler and closes channel
+	s.flush()
+	close(s.SpoolChan)
```
urso: be careful about closing a channel in worker routine. All potential senders must be stopped before.

crawler/registrar.go
```diff
@@ -16,7 +16,9 @@ type Registrar struct {
 	// Map with all file paths inside and the corresponding state
 	State map[string]*FileState
 	// Channel used by the prospector and crawler to send FileStates to be persisted
-	Persist chan *input.FileState
+	Persist       chan *input.FileState
+	running       bool
+	RegistrarChan chan []*FileEvent
```
urso: RegistrarChan?

beat/spooler.go
ruflin: That is on the todo list: https://github.com/elastic/filebeat/pull/75/files#diff-8cbd024a06b941eb89f24b91f72757dcR81 But it needs some more refactoring. The current state is the same as before, just more "local".

ruflin:
@urso Channels renamed.

crawler/registrar.go
```diff
@@ -77,10 +91,19 @@ func (r *Registrar) WriteState(input chan []*FileEvent) {
...
+
+		if !r.running {
+			break
+		}
 	}
 	logp.Debug("registrar", "Ending Registrar")
 }
 
+func (r *Registrar) Stop() {
+	r.running = false
```
driskell: Data race? :confused:

ruflin: @urso was also pointing in this direction. Now that I look at it I think the channel should be closed at the end of Run() and not in Stop(). 

@driskell Were you spotting something specific? My bigger goal is to make a "requested" shutdown more predictable especially for testing purpose.

driskell: `r.running` is being read from another routine whilst this routine is modifying it, which is racy behaviour and could trigger funky bugs where you set it to false but actually the other routine doesn't see it because it's running on a different CPU.

For Registrar I think best to use Cond or WaitGroup. Close the registrar channel from Publisher when you know there's nothing left to send. If you have multiple senders, reference count the channel, closing once all senders are finished. Otherwise you could shutdown between receiving ACK and saving the ACK to registry. This is how Log Courier handles it. Worst case you have to wait for network timeout before shutdown completes.

crawler/registrar.go
```diff
@@ -77,10 +91,19 @@ func (r *Registrar) WriteState(input chan []*FileEvent) {
 			// REVU: but we should panic, or something, right?
 			logp.Err("Update of registry returned error: %v. Continuing..", e)
 		}
+
+		if !r.running {
+			break
+		}
```
urso: why no check before writeRegistry. When function quits, writeRegistry will be called once again.

urso: plus this check is redundant to channel being closed in for-loop

crawler/registrar.go
urso: r.running is not really required. Once the channel is closed, the loop on channel will quit.

Use of r.running looks like some micro-optimization not really doing any harm. Depending on processor-type the event of setting r.running to false will not be seen by other CPU at all until next value is read from closed channel (which uses mutexes -> memory barriers included). But the fact the channel was closed will break the outer loop anyways.

More important is the subsystem startup/shutdown order in filebeat. Potential senders must be fully stopped before closing this channel. Potential consumers should be started before producers. Waiting for subsystem shutdown via sync.WaitGroup is a pretty simple/idiomatic solution. Ref-counting channels or using circuit-breakers with atomic bool might be another option complicating matters even more (adding more shutdown logic). For simplicity I'd prefer the Pipelined-susbsytems point of view (if possible).

ruflin: @urso @driskell Thanks for the inputs. I opened an issue so we can resolve this: https://github.com/elastic/filebeat/issues/77

PR #83, opened 2015-10-13
Add a system test for tailing a file

tsg:
I have a suspicion this will fail on Windows, so it's mostly a test.

### DIFF
--- a/tests/system/test_crawler.py
+++ b/tests/system/test_crawler.py
@@ -296,3 +296,43 @@ class Test(TestCase):
         # Check that output file has the same number of lines as the log file
         output = self.read_output()
         assert len(output) == 3
+
+    def test_new_line_on_open_file(self):
+        """
+        Checks that filebeat follows future writes to the same
+        file. Same as the test_new_line_on_existing_file but this
+        time keep the file open and just flush it.
+        """
+
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/log/*"
+        )
+        os.mkdir(self.working_dir + "/log/")
+
+        testfile = self.working_dir + "/log/test.log"
+        with open(testfile, 'w') as f:
+            f.write("hello world\n")
+            f.flush()
+
+            filebeat = self.start_filebeat()
+
+            self.wait_until(
+                lambda: self.log_contains(
+                    "Registrar: processing 1 events"),
+                max_timeout=15)
+
+            # now write another line
+            f.write("hello world 1\n")
+            f.write("hello world 2\n")
+            f.flush()
+
+            self.wait_until(
+                lambda: self.log_contains(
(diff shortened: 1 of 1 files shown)


tsg:
That works, but somehow I get problems doing the same thing manually on a machine. Still looking at it.

tsg:
Just to have it documented. While filebeat is reading `c:\logs\test.log`, the following Powershell command fails:

<pre>
PS C:\Windows\system32> "mmmmmmm" | out-file -FilePath c:\logs\test.log  -Append
out-file : The process cannot access the file 'C:\logs\test.log' because it is being used by another process.
At line:1 char:13
+ "mmmmmmm" | out-file -FilePath c:\logs\test.log  -Append
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OpenError: (:) [Out-File], IOException
    + FullyQualifiedErrorId : FileOpenFailure,Microsoft.PowerShell.Commands.OutFileCommand
                        
</pre>

However, adding the `-Force` parameter makes it work:

<pre>
PS C:\Windows\system32> "mmmmmmm" | out-file -FilePath c:\logs\test.log  -Append -Force
</pre>

The MSDN [documentation](https://technet.microsoft.com/en-us/library/hh849882.aspx) on `-Force` says that it allows "the cmdlet to overwrite an existing read-only file". But the error above is not about read-only and `Get-ItemProperty` also seems to indicate that the file is not read-only:

<pre>
PS C:\Windows\system32> Get-ItemProperty C:\logs\test.log -name IsReadOnly
IsReadOnly   : False
PSPath       : Microsoft.PowerShell.Core\FileSystem::C:\logs\test.log
PSParentPath : Microsoft.PowerShell.Core\FileSystem::C:\logs
PSChildName  : test.log
PSDrive      : C
PSProvider   : Microsoft.PowerShell.Core\FileSystem
</pre>

As the test from the pull request passes on Windows, I think it is not an issue with normal processes writing logs, but I wanted to have this documented in case a user is hitting something similar.

tsg:
Added another test. Provided that travis and appveyor agree, they can be merged.

tests/system/test_crawler.py
```diff
@@ -257,3 +258,81 @@ def test_file_disappear_appear(self):
...
+        testfile = self.working_dir + "/log/test.log"
+        with open(testfile, 'w') as f:
+            f.write("hello world\n")
+
+        filebeat = self.start_filebeat()
+
+        self.wait_until(
+            lambda: self.log_contains(
+                "Registrar: processing 1 events"),
+            max_timeout=15)
```
urso: can we check the filebeat output file instead of the log? Relying on filebeat debug output doesn't feel nice.

Issue #82, opened 2015-10-13
Check where the .filebeat registry file is created when installed from pacakges


tsg:
Posted so I don't forget about it. Writing the registry file in the CWD would not be ideal because that is set to the root folder by default. Somewhere in /var might be better.

tsg:
On a Windows 2012 system, the file is created at `c:\Windows\SysWOW64\.filebeat`.

tsg:
Yeah, it's `/.filebeat` on a Debian system.

I guess the Windows one is OK (don't really know what the best practices are), but the Linux one seems wrong to me, so we should discuss if we want to change it. Maybe `/var/data/filebeat` or something like that? I think but am not sure that the Logstash-Forwarder packages set it somewhere in `/opt`.

ruflin:
@tsg `/.filebat` also seems quite wrong to me. Could we set it by default in the same directory as the executable is? This should also work on Windows.

tsg:
That would be `/usr/bin`, not sure it's much better. I'm thinking to have the Linux part of the `install_cfg` target modify the configuration file to have `registryfile: /var/lib/filebeat/.filebeat`. In which case we'd also need to make sure Filebeat creates that folder if it doesn't exit.

urso:
+1 for /var/lib/filebeat on unix system. But where to put on windows?

tsg:
We put the logs under `c:\ProgramData\filebeat\Logs`, so something like `c:\ProgramData\filebeat\.filebeat` or `c:\ProgramData\filebeat\registry.json` sound good to me. Then again, people are less likely to mess with it if it's in the system folder.

urso:
+1

do we protect the file from being used by multiple filebeat instances?

tyrken:
+1 for C:\ProgramData\filebeat\registry.json on Windows - keep the system folder clear & dont write to C:\ProgramFiles outside installers.

Does it have to be ".filebeat" on Linux? It's not really a hidden file or per-user profile, more a db in the right place (/var/lib/filebeat).

ruflin:
In case it is not a hidden file, I think it makes sense to add .json to make clear the content type. The main reason to make it hidden is that it is not "removed" accidentally, but in case it is hidden in the system folders, this is not need.

@tsg Should we make the default perhaps also platfrom specific?

urso:
please do not store hidden files in /var/lib/whatever. If an admin removes this file it's his problem.

tsg:
[...]

PR #91, opened 2015-10-15
Custom registry file

tsg:
Modify the default configuration file to place the registry file on a different location depending on the platform, as requested by #82:

* On windows, `c:\ProgramData\Filebeat\registry`
* On Debian/Redhat systems if installed from pacakge, `/var/lib/filebeat/registry`.
* When downloading a tar/tgz/zip leave the default of using the current working dir

### DIFF
--- a/crawler/registrar.go
+++ b/crawler/registrar.go
@@ -2,7 +2,9 @@ package crawler
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
+	"path/filepath"
 
 	cfg "github.com/elastic/filebeat/config"
 	"github.com/elastic/filebeat/input"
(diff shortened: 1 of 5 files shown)


crawler/registrar.go
```diff
@@ -42,7 +44,19 @@ func (r *Registrar) Init() {
 		r.registryFile = cfg.DefaultRegistryFile
 	}
 
+	// Make sure the directory where we store the registryFile exists
+	absPath, err := filepath.Abs(r.registryFile)
+	if err != nil {
+		return fmt.Errorf("Failed to get the absolute path for %s: %v", r.registryFile, err)
+	}
+	err = os.MkdirAll(filepath.Dir(absPath), 0755)
```
ruflin: Should we really have the binary creating this path? I think this should be handled by the installer.

tsg: I'd like to keep the beats-packer free of Beat specific logic as much as possible, and this is something that I know it works on all platforms we support. We do the same for the log files creation in libbeat/logp.

ruflin: Ok. We can check if there is a better solution in the future. Will merge it.

PR #94, opened 2015-10-16
Rename all camelCase variables to snake_case

ruflin:
This is to adhere to the standard elasticsearch is using

### DIFF
--- a/beat/filebeat.go
+++ b/beat/filebeat.go
@@ -35,7 +35,7 @@ func (fb *Filebeat) Config(b *beat.Beat) error {
 		return fmt.Errorf("Error reading config file: %v", err)
 	}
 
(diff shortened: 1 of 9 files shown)


nik9000:
> This is to to adhere to the standard elasticsearch is using

Its called [snake_case](https://en.wikipedia.org/wiki/Snake_case).

nik9000:
Err, I call it snake_case. And so does that article. I don't mean to speak for everyone on such a controversial subject.

ruflin:
@nik9000 That is the word I was looking for when i wrote the commit message. Thanks. Title renamed.

monicasarbu:
Great. I like more snack_case. I think it fits better with Filebeat and Beats in general.

PR #99, opened 2015-10-17
Fix the install-cfg target

tsg:
There were several issues:
 * $ was a poor choice of separating character for the sed command
 * c:\program.. requires escaping in yaml because of the :
 * the default PREFIX was missing
 * s/windows/win in one place


ruflin:
Merged. OS X tests seem to be stuck currently.

PR #100, opened 2015-10-18
Experiment on the flaky registrar tests.

tsg:
I cannot get the Windows flaky tests to fail on my Windows
machine. There's a fairly theoretical race between the log file
being written and the kill command (that doesn't do any cleanup
on windows), so adding a condition to only do the kill after the
file has reached the FS.

Even if not successful, this should give us further info on what's
wrong.


tsg:
I ran Appveyor 5 times on this PR without failures. I think this is a solution for the flaky tests we've seen lately. Maybe not the cleanest solution, but it's important to have non-flaky tests as we try to get the release out.

PR #101, opened 2015-10-19
Feature/file encoding

urso:
Configureable file encoding. If not configured, file is read as is.


config/config.go
```diff
@@ -37,14 +37,19 @@ type FilebeatConfig struct {
...
 	IgnoreOlder           string `yaml:"ignore_older"`
 	IgnoreOlderDuration   time.Duration
 	ScanFrequency         string `yaml:"scan_frequency"`
 	ScanFrequencyDuration time.Duration
-	HarvesterBufferSize   int  `yaml:"harvester_buffer_size"`
-	TailOnRotate          bool `yaml:"tail_on_rotate"`
+	Harvester             HarvesterConfig `yaml:",inline"`
+}
+
+type HarvesterConfig struct {
```
urso: separated harvester config to be passed to NewHarvester to create consistent harvesters. E.g. TailOnRotate and BufferSize have been applied to stdin only.

harvester/harvester.go
```diff
@@ -10,11 +10,11 @@ import (
 type Harvester struct {
 	Path             string /* the file path to harvest */
 	ProspectorConfig config.ProspectorConfig
```
urso: ProspectorConfig still required for file 'timeout' on old files.

config/config.go
```diff
@@ -37,14 +37,19 @@ type FilebeatConfig struct {
 
 type ProspectorConfig struct {
 	Paths                 []string
-	Fields                map[string]string
```
ruflin: So far fields were defined per prospector, which I think makes sense. Like this multiple paths can have the same fields. In case every path should have different fields, multiple prospectors can be defined. Having fields in Prospector, allows the full flexibility.

config/config.go
```diff
@@ -37,14 +37,19 @@ type FilebeatConfig struct {
...
 	Input                 string
 	IgnoreOlder           string `yaml:"ignore_older"`
 	IgnoreOlderDuration   time.Duration
 	ScanFrequency         string `yaml:"scan_frequency"`
 	ScanFrequencyDuration time.Duration
-	HarvesterBufferSize   int  `yaml:"harvester_buffer_size"`
-	TailOnRotate          bool `yaml:"tail_on_rotate"`
+	Harvester             HarvesterConfig `yaml:",inline"`
```
ruflin: Same comment as I made for Fields above, applies for TailOnRotate and HarvesterBufferSize. I think these should stay in the ProspectorConfig.

ruflin:
In case we change configs from Prospector to Harvester, the default configs also have to be changed and the docs. But as mentioned above, I don't think we should change it.

config/config.go
tsg: If I understand correctly, the Fields are still configured per prospector, because HarversterConfig is "inlined" in ProspectorConfig. Or did you mean something else?

tsg:
Nice work @urso, LGTM.

monicasarbu:
@urso Great to have the encoding configured per Prospector. Thanks for making it for beta4.

config/config.go
ruflin: @tsg You are right. I totally missed the inline part. 
@urso Please ignore my comments.

monicasarbu:
@ruflin sorry, I might have pushed the merge button too early. If you still have concerns, I can revert the merge.

docs/configuration.asciidoc
```diff
@@ -105,6 +105,13 @@ filebeat:
       # Always tail on log rotation. Disabled by default
       # Note: This may skip entries
       #tail_on_rotate: false
+
+      # Configure the file encoding for reading file with international characters
+      # supported encodings:
+      #   plain, utf-8, utf-16be-bom, utf-16be, utf-16le, big5, gb18030, gbk, hzgb2312,
+      #   euckr, eucjp, iso2022jp, shiftjis, iso8859-63, iso8859-6i, iso8859-8e,
+      #   iso8859-8i
+      #encoding: utf-16
```
ruflin: Did you choose here a different one then in the default config on purpose? I suggest to also use here "plain" as this is the default, right?

ruflin:
@monicasarbu All good, @tsg solved my issue. I would still suggest to adjust the docs, but that can be done as part of an other pull request.

PR #104, opened 2015-10-19
Refresh filebeat.yml with changes from libbeat.yml.

andrewkroh:
Closes elastic/libbeat#160

### DIFF
--- a/etc/filebeat.yml
+++ b/etc/filebeat.yml
@@ -68,6 +68,7 @@ filebeat:
     #  type: stdin
     #  paths:
     #    - "-"
+
   # General filebeat configuration options
   #
   # Event count spool threshold - forces network flush if exceeded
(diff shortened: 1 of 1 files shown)


etc/filebeat.yml
```diff
@@ -89,54 +90,169 @@ filebeat:
...
- # Uncomment the following if you want to ignore transactions created
- # by the server on which the shipper is installed. This option is useful
- # to remove duplicates if shippers are installed on multiple servers.
- # ignore_outgoing: true
-
-############################# Output ############################################
+############################# Output ##########################################
 
-# Configure what outputs to use when sending the data collected by filebeat.
+# Configure what outputs to use when sending the data collected by the beat.
```
urso: s/the beat/filebeat/ ?

etc/filebeat.yml
```diff
@@ -89,54 +90,169 @@ filebeat:
...
+  #tags: ["service-X", "web-tier"]
+
+  # Uncomment the following if you want to ignore transactions created
+  # by the server on which the shipper is installed. This option is useful
+  # to remove duplicates if shippers are installed on multiple servers.
+  #ignore_outgoing: true
+
+
+############################# Logging #########################################
+
```
urso: logging looks pretty... scarce in comparison to other sections. Same for libbeat?

andrewkroh: Same for libbeat. So we can update it in libbeat and push the changes to each of the beats. Automation would be nice so we don't have to manually update each beat.

etc/filebeat.yml
andrewkroh: Fixed. But I missed this change in topbeat and packetbeat.

etc/filebeat.yml
monicasarbu: Created an issue about the logging section: https://github.com/elastic/libbeat/issues/173

Issue #109, opened 2015-10-20
Template should set the "text" field to analyzed


tsg:
The "text" field contains the log line and when sending directly to Elasticsearch (w/o Logstash). I think most people expect this to be analyzed in Elasticsearch so it's searchable.

ruflin:
Done in https://github.com/elastic/filebeat/pull/114

PR #114, opened 2015-10-20
Improved filebeat template

ruflin:
Related to issue https://github.com/elastic/filebeat/issues/109

### DIFF
--- a/etc/filebeat.template.json
+++ b/etc/filebeat.template.json
@@ -7,9 +7,34 @@
           "enabled": false
         }
       },
+      "dynamic_templates": [
+        {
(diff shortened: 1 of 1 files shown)


andrewkroh:
LGTM, but I think it would be good to get a second set of eyes on it.

ruflin:
@monicasarbu I think it would be good if you could have a look if this makes sense.

monicasarbu:
@ruflin The template looks good to me. Did you test it?

ruflin:
@monicasarbu I did some manual tests and I was able to search in the log part as expected.

Issue #77, opened 2015-10-07
Improve filebeat shutdown process


ruflin:
Currently the shutdown process of filebeat is not fully predictable. The implementation should be improved with WaitGroups. For more details see: https://github.com/elastic/filebeat/pull/75/files#r41322074

andrewkroh:
I had filebeat running for about ~4 hours on OSX and shut it down with Ctrl+C and it panicked with the attached [stacktrace](https://github.com/elastic/filebeat/files/16438/stacktrace.txt)

tsg:
I get the same on OS X when running it for only a few seconds. But I set it to read `/var/log` so it's about killing it while busy.

ruflin:
Shutdown process of filebeat was rewritten for 5.0.0. This issues should now not happen anymore. Closing.

PR #115, opened 2015-10-20
Fix crash of filebeat (temporary fix)

ruflin:
A proper fix will be done as part of https://github.com/elastic/filebeat/issues/77


crawler/registrar.go
```diff
@@ -116,6 +116,7 @@ func (r *Registrar) Run() {
 func (r *Registrar) Stop() {
 	r.running = false
 	close(r.Channel)
```
urso: might panic if the spooler send finished events to registrar.

crawler/registrar.go
```diff
@@ -116,6 +116,7 @@ func (r *Registrar) Run() {
...
+	r.writeRegistry()
```
urso: possible race between goroutine detecting channel being closed and us writing registry. In worst case to routines might try to write registry at the same time.

andrewkroh:
See #116.

PR #116, opened 2015-10-20
Fix/panic shutdown (temporary fix)

urso:
based on #115

Add signaling channel to registrar to signal shutdown without closing channel used by other goroutines.


andrewkroh:
LGTM.

Issue #112, opened 2015-10-20
Add etc/fields.yml file


monicasarbu:
To specify the exported fields and their types. Not too much to write in here, but rather useful for our users to see what fields are exported.

monicasarbu:
This file can be used to generate the 'Exported fields' chapter from the docs.

urso:
fix in #118 merged

PR #118, opened 2015-10-21
add etc/fields.yml

urso:


etc/fields.yml
```diff
@@ -0,0 +1,76 @@
...
+     type: string
+     required: true
+     description: >
+       File the line was read from.
+
+    - name: offset
+      type: int
+      required: false
+      description: >
+        File offset the reported line starts at.
```
urso: Offset really required. In comparison to line I don't see much value in offset.

etc/fields.yml
```diff
@@ -0,0 +1,76 @@
...
+      required: false
+      description: >
+        Contains user configurable fields.
+
+    - name: fileinfo
+      type: group
+      required: false
+      description: >
+        Operating System specific file information use to identify the source
+        file. For example on linux an inode might be reported.
```
urso: why index fileinfo in elasticsearch? Nice to have?

etc/fields.yml
```diff
@@ -0,0 +1,76 @@
+version: 1.0
+
+defaults:
+  type: string
+  required: false
+  index: not_analyzed
+  doc_values: true
+  ignore_above: 1024
+
+env:
```
urso: we report source OS? We'd like to have source OS?

urso:
WIP #112

etc/fields.yml
ruflin: @urso There is probably not too much value for the user, but it could be useful for the user. I would keep it in as nice to have. Give people data and they will do unexpected things with it :-)

etc/fields.yml
ruflin: Same as offset. Nice to have and good for debugging. In the long term this could even be used to get the last submitted state ...

etc/fields.yml
driskell: Offset is best IMHO.
Line number should be dropped - how can you calculate line number correctly when tailOnRotate=true?

I've never known anyone use line number, and I've only ever actually used offset in a sort (because time stamp can be identical for many lines and in Kibana you lose the ordering as a result)

monicasarbu:
@urso  thank you for adding the script as well. It fixes issue #121

PR #123, opened 2015-10-21
Add link to nightlies

ruflin:


tsg:
I was thinking we could go as far as [Kibana](https://github.com/elastic/kibana) and put direct links to the "latest" URLs? Would that be possible/wanted?

ruflin:
@tsg Yes, would be possible. I don't like to link to the latest "yet" as there are still issues with beats-tests and the latests naming.

tsg:
Ok, thanks, merging.

PR #125, opened 2015-10-21
Add ?pretty to getting started

nik9000:
Elasticsearch's docs suggest using `?pretty` when filing bugs to make things more readable so we may as well get people used to using it. Also it adds the new line!


monicasarbu:
Yes, indeed ?pretty is good to add in the docs. Thanks for the change.

PR #131, opened 2015-10-22
Removing 'fields' from JSON output when the field is null.

### LINKED ISSUE #126: Null fields are exported
Using the default config file of Filebeat, where there are no fields defined, the exported fields are null.
Here is the JSON:

"_source": {
    "count": 1,
    "fields": null,
    "fileinfo": {},
    "line": 27320,
    "message": "\t    497799835,",
    "offset": 2768605,
    "shipper": "mar.localdomain",
    "source": "/var/log/commerce.log",
    "timestamp": "2015-10-21T21:15:34.590Z",
    "type": "log"
  },

andrewkroh:
Also added timestamp to FileEvent so that the event can be timestamp as soon as it is read.

Closes #126

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ All documentation about Filebeat can be found here.
 ### Backward Compatibility Breaks
 
 ### Bugfixes
-- Omit 'fields' from event JSON when null.
+- Omit 'fields' from event JSON when null. #126
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All documentation about Filebeat can be found here.
 ### Backward Compatibility Breaks
 
 ### Bugfixes
+- Omit 'fields' from event JSON when null.
 
 ### Added
 
--- a/input/file.go
+++ b/input/file.go
@@ -2,7 +2,9 @@ package input
 
 import (
 	"os"
+	"time"
 
+	"github.com/elastic/libbeat/common"
 	"github.com/elastic/libbeat/logp"
 )
 
@@ -15,6 +17,7 @@ type File struct {
 
 // FileEvent is sent to the output and must contain all relevant information
 type FileEvent struct {
+	ReadTime time.Time
 	Source   *string
 	Offset   int64
 	Line     uint64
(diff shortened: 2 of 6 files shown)


input/file.go
```diff
@@ -44,6 +47,24 @@ func (f *FileEvent) GetState() *FileState {
 	return state
 }
 
+func (f *FileEvent) ToMapStr() common.MapStr {
+	event := common.MapStr{
+		"timestamp": common.Time(f.Time),
```
ruflin: perhaps we should even call this line something like lineReadTimestamp or something similar to better describe what it is.

andrewkroh: Yes, that would be a better. I will make that change.

andrewkroh: How about shortened to `ReadTime`?

ruflin: I like that. I'm thinking if we could have something generic that is add to all beats events. Like beatCreateTime. In a way every message sent is a beat and that is the creation time of the beat :-)

andrewkroh:
I renamed the field to `ReadTime` and also updated the CHANGELOG.md.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All documentation about Filebeat can be found here.
 ### Backward Compatibility Breaks
 
 ### Bugfixes
+- Omit 'fields' from event JSON when null.
```
ruflin: Pull request id?

ruflin:
@andrewkroh Can you update fields.yml with the new value?

CHANGELOG.md
andrewkroh: Added.

andrewkroh:
Regarding fields.yml, I updated the description of `timestamp`. Maybe we were talking about different things. Did you mean to rename the `timestamp` field present in the JSON? I was talking about renaming the `FileEvent.Time` field.

input/file.go
```diff
@@ -44,6 +47,24 @@ func (f *FileEvent) GetState() *FileState {
 	return state
 }
 
+func (f *FileEvent) ToMapStr() common.MapStr {
```
urso: With all fields of FileEvent being visible I don't feel like this requires a method, but a function

ruflin: @urso It is definitively not required, but as it is fully related to FileEvent I actually quite like that it is a method.

input/file.go
```diff
@@ -44,6 +47,24 @@ func (f *FileEvent) GetState() *FileState {
...
+	event := common.MapStr{
+		"timestamp": common.Time(f.ReadTime),
+		"source":    f.Source,
+		"offset":    f.Offset,
+		"line":      f.Line,
+		"message":   f.Text,
+		"fileinfo":  f.Fileinfo,
+		"type":      "log",
```
ruflin: @andrewkroh Just realised this line here is hardcoded. Should be dynamic, I opened and issue here for it: https://github.com/elastic/filebeat/issues/133

PR #136, opened 2015-10-24
implements technical edits and adds link to logstash configuration in…

dedemorton:
Added more detail around configuration with pointers to the steps in the libbeat documentation.

### DIFF
--- a/docs/getting-started.asciidoc
+++ b/docs/getting-started.asciidoc
@@ -1,16 +1,19 @@
 [[filebeat-getting-started]]
 == Getting Started
 
-Get started with your own Filebeat setup by first:
+To get started with your own Filebeat setup, install and configure these related products:
 
- * installing Elasticsearch for storage and indexing the data. {libbeat}/getting-started.html#elasticsearch-installation[More details]
- * installing Kibana for the UI. {libbeat}/getting-started.html#kibana-installation[More details]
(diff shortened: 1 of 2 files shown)


docs/getting-started.asciidoc
```diff
@@ -39,16 +42,21 @@ curl -L -O https://download.elastic.co/beats/filebeat/filebeat-{version}-darwin.
...
-described in the next chapter.
+Before starting filebeat, you should look at the configuration options in the configuration 
+file, for example `/etc/filebeat/filebeat.yml`. For more information about these options, 
+see <<filebeat-configuration-details,Configuration>>.
 
 [[filebeat-configuration]]
-=== Configure Filebeat
+=== Configuring Filebeat
+
+Before configuring Filebeat, you need to set up Logstash to work with the Beat. For detailed
```
ruflin: @dedemorton Having this phrased like this I worry a little bit that most users expect that Logstash is required and not optional for Filebeat. But I think this is a general problem we need to check in all the doc files so we can discuss this later.

dedemorton: @ruflin I agree. In general the doc could also use a better explanation of how filebeat works (either in the overview or a separate topic). That would give us a chance to talk about the various output options and also introduce some terminology (prospector, harvester, etc).

Issue #93, opened 2015-10-16
Cannot reuse the same name for a deleted file


tsg:
The system test `test_crawler.py:Test.test_file_disappear_appear` fails on Windows and it is currently skipped. I've looked into it and seems that the test fails for a valid reason.

Here is the problematic situation:

 * Filebeat has a log file opened
 * Another application deletes the log file (e.g. for log rotation). This is allowed by Filebeat because it opens it with the share-delete option.
 * The file is deleted but still shows up in the filesystem until all other applications close it. In our case, Filebeat still has the file opened (a harvester for it is still running).
 * While in this deleted-but-not-quite state, calling stat on the file returns errors and, more importantly, another file with the same name cannot be created.

The test tries to create a file with the same name, so that fails. I'm afraid this situation can appear in common log rotation strategies, therefore the beta4 label.

urso:
Would creating (should be supported by NTFS) hard links for log rotates files help. Using hard links we can link the file into a temporary folder and scan until original file is gone.

tsg:
Posting what I tried already here because I feel stuck. Warning, pretty raw thoughts ahead.

I tried the following two ideas:

* Replace `h.file.Stat()` with `os.Stat(h.file.Name())` [here](https://github.com/elastic/filebeat/blob/master/harvester/log.go#L219). The idea is that in the situation described above the first one still returns something while the second one returns an error. In case of errors in this point, the Harverster releases the file which is what we want. This makes the test work on Windows, but the same test fails/becomes flaky on Unix.

* In the same place, leave the `h.file.Stat()` but add also a second call to `os.Stat()` so we check the result from both and compare. I.e.:

<pre>
@@ -213,6 +213,7 @@ func lineEndingChars(line []byte) int {

 // Handles error during reading file. If EOF and nothing special, exit without errors
 func (h *Harvester) handleReadlineError(lastTimeRead time.Time, err error) error {
        if err == io.EOF {
                // timed out waiting for data, got eof.
                // Check to see if the file was truncated
@@ -224,6 +225,15 @@ func (h *Harvester) handleReadlineError(lastTimeRead time.Time, err error) error
                        return statErr
                }

+               // On Windows, if h.File.Stat() succeeds and os.Stat() fails on the
+               // same file name, it usually means that the file was removed but it is
+               // still open by us. Make sure we return an error so that the file is
+               // closed.
+               if _, osStatErr := os.Stat(h.file.Name()); osStatErr != nil {
[...]

PR #146, opened 2015-10-27
Fix windows issue by making special check for windows

ruflin:
See https://github.com/elastic/filebeat/issues/93

Remove skipping tests
Introduce running variable to directly stop after file removal.

* [x] Discuss configuration options
  * What should be configurable?
  * What are the default options?
  * Variables to discuss:
    * backoff
       * maxBackoff
       * backoffFactor
    * partialLineWaitingTime
    * forceCloseFilesWindows
* [x] introduce config variables
* [x] Update CHANGELOG.md

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,14 +12,19 @@ import (
 
 // Defaults for config variables which are not set
 const (
-	DefaultRegistryFile                      = ".filebeat"
-	DefaultIgnoreOlderDuration time.Duration = 10 * time.Minute
-	DefaultScanFrequency       time.Duration = 10 * time.Second
-	DefaultSpoolSize           uint64        = 1024
-	DefaultIdleTimeout         time.Duration = 5 * time.Second
-	DefaultHarvesterBufferSize int           = 16 << 10 // 16384
-	DefaultDocumentType                      = "log"
-	DefaultTailFiles                         = false
+	DefaultRegistryFile                         = ".filebeat"
+	DefaultIgnoreOlderDuration    time.Duration = 10 * time.Minute
+	DefaultScanFrequency          time.Duration = 10 * time.Second
+	DefaultSpoolSize              uint64        = 1024
+	DefaultIdleTimeout            time.Duration = 5 * time.Second
+	DefaultHarvesterBufferSize    int           = 16 << 10 // 16384
+	DefaultDocumentType                         = "log"
+	DefaultTailFiles                            = false
+	DefaultBackoff                              = 1 * time.Second
+	DefaultBackoffFactor                        = 2
+	DefaultMaxBackoff                           = 10 * time.Second
+	DefaultPartialLineWaiting                   = 5 * time.Second
+	DefaultForceCloseWindowsFiles               = false
 )
 
 type Config struct {
@@ -46,12 +51,20 @@ type ProspectorConfig struct {
 }
 
 type HarvesterConfig struct {
-	InputType    string `yaml:"input_type"`
-	Fields       map[string]string
-	BufferSize   int    `yaml:"harvester_buffer_size"`
-	TailFiles    bool   `yaml:"tail_files"`
-	Encoding     string `yaml:"encoding"`
-	DocumentType string `yaml:"document_type"`
+	InputType                  string `yaml:"input_type"`
+	Fields                     map[string]string
+	BufferSize                 int    `yaml:"harvester_buffer_size"`
+	TailFiles                  bool   `yaml:"tail_files"`
+	Encoding                   string `yaml:"encoding"`
+	DocumentType               string `yaml:"document_type"`
+	Backoff                    string `yaml:"backoff"`
+	BackoffDuration            time.Duration
+	BackoffFactor              int    `yaml:"backoff_factor"`
+	MaxBackoff                 string `yaml:"max_backoff"`
+	MaxBackoffDuration         time.Duration
+	PartialLineWaiting         string `yaml:"partial_line_wating"`
+	PartialLineWaitingDuration time.Duration
+	ForceCloseWindowsFiles     bool `yaml:"force_close_windows_files"`
 }
 
 // getConfigFiles returns list of config files.
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -142,6 +142,52 @@ Setting tail_files to true means filebeat starts readding new files at the end
 instead of the beginning. If this is used in combination with log rotation
 this can mean that the first entries of a new file are skipped.
 
+===== backoff
+
+Backoff values define how agressively filebeat crawls new files for updates
+The default values can be used in most cases. Backoff defines how long it is waited
+to check a file again after EOF is reached. Default is 1s which means the file
+is checked every second if new lines were added. This leads to a near real time crawling.
+Every time a new line appears, backoff is reset to the initial value.
+Default: 1s
+
+===== max_backoff
+
+Max backoff defines what the maximum waiting time is. After having backed off multiple times
+from checking the files, the waiting time will never exceed max_backoff idenependent of the
+backoff factor. Having it set to 10s means in the worst case a new line can be added to a log
+file after having backed off multiple times, it takes a maximum of 10s to read the new line.
+Default: 10s
+
+===== backoff_factor
+
+The backoff factor defines how fast the waiting time is increased. The bigger the backoff factor,
+the faster the max_backoff value is reached. The backoff increments exponential.
+The minimal value allowed is 1. If this value is set to 1 it means backoff algorithm is disabled
+and the backoff value is used for waiting for new lines.
+The backoff value will be multiplied each time with the backoff_factor until max_backoff is reached.
+Default: 2
+
+===== partial_line_waiting
+
+Defines the time on how long the harvester will wait for a line to be completed.
+Sometimes a lines it not completely written when checked by filebeat. Filebeat
+will wait for the time defined below so the system can complete the line.
+In case the line is not completed in this time, the line will be skipped.
+Default: 5s
+
+===== force_close_windows_files
+
+This option closes a file on windows, as soon as the file name changes.
+This config option is windows only. Filebeat keeps the files it's reading open. This can cause
+issues when the file is removed, as the file will not be fully removed until also filebeat closes
+the reading. Filebeat closes the file handler after ignore_older. During this time no new file with the
+same name can be created. Turning this feature on the other hand can lead to loss of data
+on rotate files. It can happen that after file rotation the beginning of the new
+file is skipped, as the reading starts at the end. We recommend to leave this option on false
+but lower the ignore_older value to release files faster.
+Default: false
+
 ===== spool_size
 
 Event count spool threshold - forces network flush if exceeded.
--- a/harvester/log.go
+++ b/harvester/log.go
@@ -3,9 +3,11 @@ package harvester
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"time"
 
 	"github.com/elastic/filebeat/config"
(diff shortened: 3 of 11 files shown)


ruflin:
@urso @monicasarbu @andrewkroh @tsg With this pull request several new configuration options are introduced. I would like to discuss here which of these we should make available in the config file and what default values we should select.

 * `forceCloseFilesWindows`: This variable was introduced to decide if EOF of a file is reached, if it should be closed by default or if it should be waited to expire (ignore_older). Releasing it means, file rotation works as expected. But it can happen that the file still exist and it is still written to the rotated file and data is lost. I suggest to make this variable available by default as the user should make a conscious decision about the tradeoff he makes. Not sure about the default value.

* Backoff: I introduced a backoff algorithm so a log file that is not constantly written to is not checked every second for new entries. It is currently unclear to me if the checks every second has an effect on resource usage as it could well be that the file system has all the data in cache anyways and not lots of  additional resources are used. As we want to stay "near" real time I introduced also the `maxBackoff` time and set it to 60s currently. So in the worst case if a file was not updated for a long time, it will take maximum 60s until new lines are discovered. The factor `backoffFactor` currently used is 2, means with each iteration the backoff time doubles until maxBackoff reached.
* `partialLineWaitingTime`: How long does the reader wait before a partial line is discarded. I changed to implementation so every time a new segment is added to the line, the counter is reset. The current default value is 5s, so it means if for more then 5 seconds a line is not completed, it will be dropped. This valued could also get relevant for multiline.

Some of the above backoff values could get obsolete when we can use https://github.com/go-fsnotify/fsnotify

We must decide which of the variables should be added to the config file and which are the default values.

ruflin:
@driskell Would also be interesting to hear your opinion on the above.

harvester/log.go
```diff
@@ -36,52 +38,109 @@ func NewHarvester(
...
+						return
+					}
+				}
+
+				// Wait before trying to read file which reached EOF again
+				time.Sleep(backoff)
+
+				// Increment backoff up to a minute
+				if backoff < maxBackoff {
+					backoff = backoff * backoffFactor
```
urso: small bug here.

The backof algo implements:

    backoff_i = backoff * 2^i

with backoff = 1s and maxBackoff = 60s we will see for i = 5:

    backoff_5 = backoff * 32 = 32s

and for i = 6:

    backoff_6 = backoff * 64 = 64s > maxBackoff.

we basically want

    backoff_i = min(backoff_(i-1) * backoffFactor, maxBackoff)

ruflin: @urso You are of course right. Interestingly, I fixed it already but forgot to push the last commit (see the change below). I think it does the same it is just written a little bit different.

config/config.go
```diff
@@ -12,14 +12,19 @@ import (
...
+	DefaultScanFrequency          time.Duration = 10 * time.Second
+	DefaultSpoolSize              uint64        = 1024
+	DefaultIdleTimeout            time.Duration = 5 * time.Second
+	DefaultHarvesterBufferSize    int           = 16 << 10 // 16384
+	DefaultDocumentType                         = "log"
+	DefaultTailOnRotate                         = false
+	DefaultBackoff                              = 1 * time.Second
+	DefaultBackoffFactor                        = 2
+	DefaultMaxBackoff                           = 10 * time.Second
+	DefaultPartialLineWaiting                   = 5 * time.Second
```
urso: I'd like to have partial line off. If some buffering is preventing the line from being complete I can wait a few minutes to finally have line indexed. It's more of an advanced/unsafe config option for impatient users who are willing to trade correctness for latency.

docs/configuration.asciidoc
```diff
@@ -140,6 +140,49 @@ Defines the buffer size every harvester uses when fetching the file. By default
...
+Backoff values define how agressively filebeat crawls new files for updates
+The default values can be used in most cases. Backoff defines how long it is waited
+to check a file again after EOF is reached. Default is 1s which means the file
+is checked every second if new lines were added. This leads to a near real time crawling.
+Every time a new line appears, backoff is reset to the initial value.
+Default: 1s
+
+===== max_backoff
+
+Max backoff defines what the maximum backoff time is. After having backed off multiple times
```
urso: can we somehow add 'wait' in here. reading backoff and back off makes my head spin.

docs/configuration.asciidoc
```diff
@@ -140,6 +140,49 @@ Defines the buffer size every harvester uses when fetching the file. By default
...
+from checking the files, the waiting time will never exceed max_backoff idenependent of the
+backoff factor. Having it set to 10s means in the worst case a new line can be added to a log
+file after having backed off multiple times, it takes a maximum of 10s to read the new line.
+Default: 10s
+
+===== backoff_factor
+
+The backoff factor defines how fast the algorithm backs off. The bigger the backoff factor,
```
urso: - "how fast" the waiting time is increased.
- mention exponential backoff
- min allowed value: 1 (disable backoff algorithm => backoff is used for waiting for new lines written to file)

docs/configuration.asciidoc
```diff
@@ -140,6 +140,49 @@ Defines the buffer size every harvester uses when fetching the file. By default
...
+The backoff value will be multiplied each time with the backoff_factor until max_backoff is reached
+Default: 2
+
+===== partial_line_waiting
+
+Defines the time on how long the harvester will wait for a line to be completed.
+Sometimes a lines it not completely written when checked by filebeat. Filebeat
+will wait for the time defined below so the system can complete the line.
+In case the line is not completed in this time, the line will be skipped.
+Default: 5s
```
urso: How to disable? It's ok to have this option, but by default we should try to not drop any lines.

docs/configuration.asciidoc
```diff
@@ -140,6 +140,49 @@ Defines the buffer size every harvester uses when fetching the file. By default
...
+
+===== force_close_windows_files
```
urso: maybe introduce platform specific sub-sections to config file?

    prospector:
      windows:
        force_close_files: ...

docs/configuration.asciidoc
```diff
@@ -140,6 +140,49 @@ Defines the buffer size every harvester uses when fetching the file. By default
...
+
+This option closes a file on windows, as soon as the file name changes.
+This config option is windows only. Filebeat keeps the files it's reading open. This can cause
+issues when the file is removed, as the file will not be fully removed until also filebeat closes
+the reading. Filebeat closes the file handler after ignore_older. During this time no new file with the
+same name can be created. Turning this feature on can on the other hand lead to loss of data
+on rotate files as rotate files are not followed. We recommend to leave this option on false
```
urso: on file rotation as rotated failes are not followed?

harvester/log.go
```diff
@@ -128,11 +183,12 @@ func (h *Harvester) setFileOffset() {
 	}
 }
 
-func (h *Harvester) open() *os.File {
+// open Opens the file given under h.Path and assigns the file handler to h.file
```
urso: open opens ... ?

harvester/log.go
```diff
@@ -242,36 +293,58 @@ func lineEndingChars(line []byte) int {
...
+// In case of partial lines, readLine waits for a maximum of 5 seconds for new segments to arrive.
+// This could potentialy be improved / replaced by https://github.com/elastic/libbeat/tree/master/common/streambuf
+func readLine(reader *bufio.Reader, buffer *bytes.Buffer, partialLineWaiting time.Duration) (*string, int, error) {
 
-		// This could happen if the file was removed / rotate after reading and before calling the stat function
-		if statErr != nil {
-			logp.Err("Unexpected error reading from %s; error: %s", h.Path, statErr)
-			return statErr
+	lastSegementTime := time.Now()
+	partialLine := true
```
urso: how can I disable this from config file without setting timeout to 10 years?

config/config.go
ruflin: @urso I think we can increase this value but I would not completely set it to false in case somehow a line is broken (for whatever reason), it is still skipped after some time. I suggest we set it to maxBackoff by default?

docs/configuration.asciidoc
ruflin: Rephrased to: Max backoff defines what the maximum waiting time is

docs/configuration.asciidoc
ruflin: Changed to: 

The backoff factor defines how fast the waiting time is increased. The bigger the backoff factor,
the faster the max_backoff value is reached. The backoff increments exponential.
The minimal value allowed is 1. If this value is set to 1 it means backoff algorithm is disabled
and the backoff value is used for waiting for new lines.
The backoff value will be multiplied each time with the backoff_factor until max_backoff is reached.

docs/configuration.asciidoc
ruflin: See comment above. Currently it is not possible to disable this option. It prevents us also from getting stuck with a corrupted line.

docs/configuration.asciidoc
ruflin: I would do this as soon as we have multiple platform specific configs. In my ideal work there aren't any ;-)

docs/configuration.asciidoc
ruflin: Changed to: 

This option closes a file on windows, as soon as the file name changes.
This config option is windows only. Filebeat keeps the files it's reading open. This can cause
issues when the file is removed, as the file will not be fully removed until also filebeat closes
the reading. Filebeat closes the file handler after ignore_older. During this time no new file with the
same name can be created. Turning this feature on the other hand can lead to loss of data
on rotate files. It can happen that after file rotation the beginning of the new
file is skipped, as the reading starts at the end. We recommend to leave this option on false
but lower the ignore_older value to release files faster.

harvester/log.go
ruflin: fixed: // open does open the file given under h.Path and assigns the file handler to h.file

harvester/log.go
ruflin: Not possible at the moment with the current implementation.

Side note: This variable should probably be called isPartialLine as it is only a "check" variable.

ruflin: I renamed it to isPartialLine to make it more clear

ruflin:
@urso feedback implemented.

urso:
LGTM

docs/configuration.asciidoc
driskell: I don't know what you mean by "corrupted line". How is a line corrupted? And why would it matter to filebeat?

I can only think of one situation and that is where there is no line ending, but if we drop corrupt lines that's the same as buffering them from a user perspective, you just don't lose it if a line ending appears. Surely not losing it is better?

ruflin: Yep, probably a corrupted line would mean a line we can't detect that is one (line ending we didn't know about). This is the line we are talking about: https://github.com/elastic/filebeat/pull/146/files#diff-c1434f19d1ba7920b4f4a6aa9c5e66b9R329

@driskell @urso If I understand you right you would suggest the following:
* Return what we have instead of dropping the line

In this case, would the waiting be ok or would you still suggest to wait as long as possible?

docs/configuration.asciidoc
driskell: Why not make it for all platforms?
I know it's intended for Windows users but maybe it serves a purpose on Linux too if people want files closed quickly.

I think only thing to be sure about is how the resume after close works as I remember having some pretty funky edge cases with missing lines caused by racy behaviour where new lines would appear at the exact moment after the Read returned EOF, and at an exact moment in Prospector. :-(

harvester/log.go
```diff
@@ -36,54 +38,107 @@ func NewHarvester(
...
+		return
+	}
+
 	// On completion, push offset so we can continue where we left off if we relaunch on the same file
 	defer func() {
 		h.FinishChan <- h.Offset
 	}()
 
-	var line uint64 = 0 // Ask registrar about the line number
+	var line uint64 = 0
```
driskell: This is generally always wrong when you restart forwarder as even though it resumed a file it would set the line number to 0. So I don't know if you want to remove this field?

driskell:
Dropped a couple notes in :-) I didn't check it in detail though - looks like you all have it covered! So based on that LGTM!

docs/configuration.asciidoc
ruflin: Interesting idea. I'm sure someone will also request it for the other platforms. It could be interesting especially in high load environments where the log files are rotate often. 

I'm curious what edge cases will pop up here :-( Did you find a specific fix for it?

ruflin: @urso @tsg what you think about making the config for all platforms? Would make the code cleaner.

docs/configuration.asciidoc
driskell: Return what you have is good.
Only thing that comes to mind is if disk freezes for too long in between writing blocks it could split a line. It'll be a really rare edge case though, especially if you wait minutes. But at least nothing is technically lost - though in a large system a split line has no timestamp so could become a needle in a haystack that's never seen.

docs/configuration.asciidoc
driskell: Edge case is to do with coordinating the stop and resume.

File hits EOF just before the file is updated
Prospector scans it with the NEW mod time, and moves on thinking it still running
Harvester pushes the final offset to the channel
Prospector scans again, sees harvester finished, but doesn't see mod time changed from last scan so doesn't restart the harvester

I think that's the only one. Likely rare. Slow VMs don't help the situation though! Only affects the last line of a file too as subsequent writes trigger it. Just something to consider when increasing the rate at which files are closed/resumed.

harvester/log.go
ruflin: Good point. Either we find a way to set it to the correct line or we remove it. At the same time for log files it doesn't really provide any value except for debugging.

ruflin:
@driskell Thanks for the feedback. I openend here a follow up issue to discuss the open points: https://github.com/elastic/filebeat/issues/216

PR #150, opened 2015-10-27
Fix #149: Remove espacing of path

### LINKED ISSUE #149: Package config file for Windows is invalid YAML
The current config file which is downloaded with the beta4 windows packages is invalid. The reason is the escaping of f in the registry_path which leads to a special character. The build script must be adapted:

https://github.com/elastic/filebeat/blob/master/Makefile#L86

screen shot 2015-10-27 at 15 19 45

ruflin:
By removing the espacing of the F in the registry_path valid YAML is generated again

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,4 +83,4 @@ install-cfg:
 	cp etc/filebeat.yml $(PREFIX)/filebeat-darwin.yml
 	# win
 	cp etc/filebeat.yml $(PREFIX)/filebeat-win.yml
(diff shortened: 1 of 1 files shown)


ruflin:
@tsg Can you review this?
Would be nice if this would be detected by beats-tester.

andrewkroh:
LGTM

urso:
LGTM

monicasarbu:
It looks like a typo to me, merging it.

tsg:
Unfortunately beats-tester uses it's own config file, because making a find and replace in powershell turned out to be quite difficult. That and this typo are just sad :-(. Thanks for fixing.

Issue #133, opened 2015-10-23
Make type field dynamic based on input


ruflin:
Currently the type of an event is hardcoded to "log". This must be dynamic based on the harvester that was started in the prospector config.

This line must be changed: https://github.com/elastic/filebeat/blob/master/input/file.go#L58

urso:
@ruflin can you check whether #153 correctly addresses this ticket?

ruflin:
@urso Right direction :-)

ruflin:
Done by #153

PR #153, opened 2015-10-27
make event type configurable

urso:


config/config.go
```diff
@@ -50,6 +51,7 @@ type HarvesterConfig struct {
 	BufferSize   int    `yaml:"harvester_buffer_size"`
 	TailOnRotate bool   `yaml:"tail_on_rotate"`
 	Encoding     string `yaml:"encoding"`
+	Type         string `yaml:"event_type"`
```
andrewkroh: Consider renaming `Type` to `EventType` for consistency. The field names seem to only differ from their yaml name when underscores have been added.

ruflin: @urso In the config it is called `type`, so this mapping will not work. Also I think it is the "prospector type" and not the event type as it is a prospector config. I would stay with "type".

ruflin: I just realised that this is in HarvesterConfig. Should be in PropsectorConfig. That is the config line we are referring to: https://github.com/elastic/filebeat/blob/master/etc/filebeat.yml#L31

urso:
@ruflin so the ticket refers to the already present type field in the prospector? Was unsure about this and added event_type (also in prospector configuration section) which I think is more flexible in the end. Maybe have event_type and if event_type is not set fall back to type?

ruflin:
@urso Yes, was referring to the existing type in prospector. I think one prospector should take care of only 1 type of harvesters, that why I suggest to keep it on the prospector level.

urso:
@ruflin but depending on path you might want to configure multiple prospectors, right? Then having extra event_type can be useful

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All documentation about Filebeat can be found here.
 ### Bugfixes
 - Omit 'fields' from event JSON when null. #126
 - Make offset and line value of type long in elasticsearch template to prevent overflow #140
+- Make 'type' field in JSON event configurable. #133
```
ruflin: Suggested alternative: Introduce 'event_type' config option per prospector to define document type for event stored in elasticsearch

urso: sounds good

docs/fields.asciidoc
```diff
@@ -36,14 +36,14 @@ format: YYYY-MM-DDTHH:MM:SS.milliZ
...
-The timestamp when the log line is published. The precision is in milliseconds. The timezone is UTC.
+The timestamp when the log line was read. The precision is in milliseconds. The timezone is UTC.
 
 
 ==== type
 
 required: True
 
-Set to "log" to telling the event contains content read from log files.
+Name of the log event. Set to the prospectors event_type configuration option.
```
ruflin: This is not correct, right?

type: Type of the prospector and the harvestors to start. This can be "log" or "stdin". Default is "log".

etc/fields.yml
```diff
@@ -28,7 +28,7 @@ env:
 
     - name: type
       description: >
-        Set to "log" to telling the event contains content read from log files.
+        Name of the log event. Set to the prospectors event_type configuration option.
```
ruflin: See above

docs/fields.asciidoc
urso: default is DefaultEventType. This exported field matches 'event_type' configuration

etc/filebeat.yml
```diff
@@ -40,6 +40,9 @@ filebeat:
       # Time strings like 2h (2 hours), 5m (5 minutes) can be used.
       #ignore_older:
 
+      # Event type to be published in 'type' field. Default: log
```
ruflin: Event type defines the document type these entries should be stored in Elasticsearch. Default: log

docs/fields.asciidoc
urso: This is the exported field used for document type in elasticsearch by default

etc/fields.yml
urso: yep, fields.asciidoc is generated from fields.yml.

urso:
resolves #133

docs/fields.asciidoc
ruflin: But is this field still exported when it is used as document_type? Will it exist in es as field?

ruflin:
@urso With the current change it looks like the "prospector type" will not be exported at all anymore. I think this would be still worth to export.

urso:
@ruflin suggestion? Add 'prospector_type' field to published event?

ruflin:
@urso `input_type`? I'm thinking about renaming `event_type` to `document_type` in the config to make it more obvious. And make the current type to `input_type`. Otherwise I already see the confusion happening as type was used before for document_type in logstash.

urso:
+1 for renaming event_type to document_type
+1 for using input_type. We should also rename 'type' to 'input_type' in config file to make it consistent.

ruflin:
@urso agree. It will break BC with beta4 in case someone was using stdin type. We need to update the docs accordingly.

docs/fields.asciidoc
urso: I think so. But this depends on libbeat.

ruflin: Agree

urso:
don't think it breaks BC, as default value in event_type is still 'log'.

docs/configuration.asciidoc
```diff
@@ -95,6 +95,9 @@ It has two options:
     * log:  Reads every line of the log file (default)
```
ruflin: On line 91, it should now also be 'input_type' Unfortunately github only allows me to comment here.

ruflin:
@urso LGTM. One last comment above and can you squash afterwards?

urso:
updates and squashed. Hope I didn't miss anything else.

ruflin:
@urso There is one more in the test config file. But I will add some rudementary tests for the above and will fix it directly.

owulff:
Will the attribute name 'type' of @metadata sent to logstash change or just the value.
    "@metadata": {
      "beat": "<beat>",
      "type": "<event type>"
    }

ruflin:
@owulff Based on https://github.com/elastic/filebeat/pull/153/files#diff-05dec33cd24c1d98f182a2a941b53f79R60, only the value will change. Means it will depend on document_type instead of input_type in the config. Default will still be log if not set. A new field input_type is now also sent to logstash.

Does this work in your use case?

owulff:
:+1:

PR #164, opened 2015-10-30
Set ignore_older to 10m by default

ruflin:
This will close open files much faster and leads to less open files.
As soon as the prospector scans for new files again and the file has changed,
it will be picked up again.


monicasarbu:
LGTM

PR #168, opened 2015-10-30
Rename timestamp with @timestamp

monicasarbu:
* [x] rename `timestamp` with `@timestamp` in the code and tests
* [x] update godeps
* [x] update fields.yml 
* [x] update docs
* [x] update template
* [x] update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All documentation about Filebeat can be found here.
 - Add 'input_type' field to published events reporting the prospector type being used. #133
 
 ### Added
+- Rename the timestamp field with @timestamp #168
 
 ### Improvements
 - All Godeps dependencies were updated to master on 2015-10-21 [#122]
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -7,31 +7,38 @@
 	"Deps": [
 		{
 			"ImportPath": "github.com/elastic/libbeat/beat",
-			"Rev": "3989f8f01bb335d756bad200f8b3b1fa0ae941db"
+			"Comment": "v1.0.0-beta4-42-g00f00f9",
+			"Rev": "00f00f9e991b32db10ce2a10f825525b35e38a52"
 		},
 		{
 			"ImportPath": "github.com/elastic/libbeat/cfgfile",
-			"Rev": "3989f8f01bb335d756bad200f8b3b1fa0ae941db"
+			"Comment": "v1.0.0-beta4-42-g00f00f9",
+			"Rev": "00f00f9e991b32db10ce2a10f825525b35e38a52"
 		},
 		{
 			"ImportPath": "github.com/elastic/libbeat/common",
-			"Rev": "3989f8f01bb335d756bad200f8b3b1fa0ae941db"
+			"Comment": "v1.0.0-beta4-42-g00f00f9",
+			"Rev": "00f00f9e991b32db10ce2a10f825525b35e38a52"
 		},
 		{
 			"ImportPath": "github.com/elastic/libbeat/logp",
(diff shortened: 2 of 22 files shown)


urso:
generating  fields.asciidoc missing

DanielLaberge:
Since this was merged, I'm getting the following error in logstash-input-beats: 
```
"Beats input: unhandled exception", :exception=>#<TypeError: The field '@timestamp' must be a (LogStash::Timestamp, not a String (2015-11-06T22:07:14.766Z)>, :backtrace=>["/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.0.0.rc1-java/lib/logstash/event.rb:138:in `[]='", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-0.9.2/lib/logstash/inputs/beats.rb:138:in `create_event'", ....
```
I'm using the Logstash:2 docker image, which seems to install **version 0.9.2** of logstash-input-beats.

I imagine once the Docker image installs the latest version of this plugin, my error should be resolved?

If so, could the the image be updated? Or can I do anything in my logstash.conf to work around this?

urso:
@DanielLaberge 
which docker image? I don't think we've any official docker images. But yes, the default beats plugin shipped with Logstash 2.0.0 is old and has some bugs. current version is 0.9.5. In this case you case use your own Dockerfile deriving from Logstash:2.0.0 and update the plugin. Some of our tests still use logstash 1.5.4 and install plugin during build phase: see https://github.com/elastic/libbeat/tree/master/scripts/docker/logstash . Fortunately logstash is pretty self contained, thus building your own Docker image is pretty easy.

DanielLaberge:
This one: https://hub.docker.com/_/logstash/

urso:
Ah yes, used to use this one too until switching to my own Dockerfile for testing with Logstash. Don't think this is an official image by logstash team. Reading the Dockerfile it just get's you Logstash 2.0.0 which still comes with the outdated plugin.

alastrange:
Getting the same problem as @DanielLaberge with logstash 2.0.0 downloaded directly from ES

```TypeError: The field '@timestamp' must be a (LogStash::Timestamp, not a String (2015-11-06T22:07:14.766Z)```

I'll try updating the plugins

ruflin:
@alastrange Here are the docs on how to update the plugin: https://www.elastic.co/guide/en/beats/libbeat/current/getting-started.html#logstash-input-update

EamonZhang:
after update is ok now

op-fasc:
this causes a problem if the timestampfield from log is also "@timestamp". The timestamp from log will be overwritten.
https://github.com/logstash-plugins/logstash-input-beats/issues/33

Is there a way to ignore the @timestamp from filebeat when using json codec?

ruflin:
@op-fasc I would actually hope that this is a problem that can be solved on the logstash side. An other solution could be to use JSON decoding directly in filebeat which comes with 5.0: https://github.com/elastic/beats/pull/1143 Then you can also select if the fields should be overwritten or not: https://github.com/elastic/beats/pull/1143/files#diff-2e6aee784a6a7119dda6783214706ae9R45

op-fasc:
Thanks for your answer. This may be a solution but if there are problems parsing a json file, this has to be handled on client side. Actually i capture all json parse errors in logstash and write them in a file to identify the bad formatted logentries. If beats has parse errors i have to chack all servers...

I came around my problem by using plain-codec for beats-input and a json-filter afterwards. So beats-input doesn't recognize "@timestamp" from the message and json-filter overwrites existing "@timestamp" from beats. Unfortunately this is a unnecessary step and it seems that json parsing with codec from beats-input works a little bit better, but it works...

ruflin:
@op-fasc In case filebeat would ship its own logs also to elasticsearch, you could check the errors in a centralised place ...

Issue #163, opened 2015-10-29
links in getting started doc point to the wrong version of the libbeat doc


dedemorton:
The More details links in the Beta4 getting started docs (https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-getting-started.html) point to the master version of libbeat rather than the current version.

dedemorton:
Fixed this problem in all beats.

PR #169, opened 2015-10-30
fixes linking for issue #163

dedemorton:


monicasarbu:
@dedemorton Sorry for the confusion, I just realized that setting the `:libbeat:` variable to /current will not work for older versions. For example if beta4 is the current then the beta3 documentation should link to beta3 and not to current (beta4). 
By changing the `:libbeat:` variable manually to point to the right version when we create a release branch we fix this issue. For example:
 * in the beta3 branch, `:libbeat:` points to /1.0.0-beta3
* in beta4 branch, `:libbeat:` points to /1.0.0-beta4
* in master, `:libbeat:` points to /master

dedemorton:
@tsg I thought the yaml file took care of setting the value for current so that we don't have to specify the version in our libbeat variable and have to update it every time we push files into master. Can you clarify?

tsg:
Yeah, I think it's better to link to a fixed version (e.g from the filebeat 1.0.0-beta4 branch we link to libbeat 1.0.0-beta4).

We change the current label to point to the latest stable release (from the `docs/conf.yaml` file indeed), so if we'd used current in our links, Filebeat 1.0.0-beta4 will point to Libbeat 1.0.0-ga once we release GA. This wouldn't be good, because the newer versions might have more configuration options documented, which wouldn't be in the software version the user tries.

dedemorton:
thanks, that makes sense now. I will change the version as Monica recommends.

PR #174, opened 2015-11-02
Add multi append system test.

ruflin:
Set idle_timeout to 0.5s to speed up system tests

### DIFF
--- a/tests/system/test_crawler.py
+++ b/tests/system/test_crawler.py
@@ -297,6 +297,52 @@ class Test(TestCase):
         output = self.read_output()
         assert len(output) == 3
 
+    def test_multiple_appends(self):
+        """
+        Test that filebeat keeps picking up new lines
+        after appending multiple times
+        """
+
+        self.render_config_template(
+            path=os.path.abspath(self.working_dir) + "/log/*"
+        )
+        os.mkdir(self.working_dir + "/log/")
+
+        testfile = self.working_dir + "/log/test.log"
+
+        filebeat = self.start_filebeat()
+
+        # Write initial file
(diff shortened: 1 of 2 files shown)


tests/system/test_crawler.py
```diff
@@ -297,6 +297,52 @@ def test_new_line_on_existing_file(self):
...
+        with open(testfile, 'w') as f:
+            f.write("hello world\n")
+            f.flush()
+
+            self.wait_until(
+                lambda: self.log_contains(
+                    "Registrar: processing 1 events"),
+                max_timeout=15)
+
+        for n in range(0, 3):
```
urso: range(0, 3) == range(3) ?

tests/system/test_crawler.py
```diff
@@ -297,6 +297,52 @@ def test_new_line_on_existing_file(self):
...
+
+                for i in range(0, 20 + n):
+                    f.write("hello world " + str(i) + " " + str(n) + "\n")
+
+                f.flush()
+
+                self.wait_until(
+                    lambda: self.log_contains(
+                        "Registrar: processing " + str(20 + n) + " events"),
+                    max_timeout=15)
```
urso: this exact string racy? What if we change some logic regarding processing and registrar => potentially flaky tests?

ruflin: Yes, but this will affect most of the tests. It could be definitively improved but at the moment it is a good way to track progress of filebeat.

tests/system/test_crawler.py
ruflin: adjusted

PR #175, opened 2015-11-02
Rename tail_on_rotate to tail_files

ruflin:


urso:
rebase required

urso:
LGTM. Still I don't like the behavior of dropping lines after rotating, as a rotated file is kind of a continuation of the log. Still tailFiles is a nice feature to have if I don't want to index already present log files when starting filebeat the first time.

ruflin:
@urso what you suggest is finding a way so we can detect if the file is an old file or a rotate one. One solution here would be to "tail" only the first time the scanning is done. From then on, all new discovred files will not be tailed.

ruflin:
Rebase done.

@urso Should we move forward here or not? On the one hand I think tail_files is better then tail_on_rotate naming wise, at the same time tail_on_rotate is more a warning on what could go wrong. The functionality currently is the same as before. I would suggest to move forward and in a next step try to prevent the tailing of rotate files.

PR #176, opened 2015-11-02
Update logstash migration TLS configuration documentation

steven-hadfield:
The migration configuration currently uses dashes for the certificate configuration instead of underscores. Additioinally, the certificate key was incorrectly specified


monicasarbu:
@steven-hadfield Thank you for the fix. Can you please sign the CLA?

steven-hadfield:
I did sign the CLA, but it seemed like it was taking a while to acknowledge it. It appears that it has gone through now.

PR #182, opened 2015-11-03
Fix #167. Reload encoding transformer at the end of each file

### LINKED ISSUE #167: Filebeat does not tail log files on Windows when encoding is set to UTF-8
When I start filebeat, it sends all new log entries to logstash as expected, **but then it stops after reaching the end of the log files and does not send new entries as they are added.** 

This is blocking me from using filebeat, any guidance would be much appreciated.

Using the latest nightly build (filebeat-nightly.151030140642-windows.zip).

Tested on Windows 8.1 Pro x64 on an NTFS file system and Windows Server 2012 R2 on NTFS with the same results.

**Filebeat log:**
````
C:\logs\filebeat>filebeat.exe -c filebeat.yml -v
geolite.go:52: WARN Couldn't load GeoIP database
publish.go:222: WARN No output is defined to store the topology. The server fields might not be filled.
publish.go:234: INFO No shipper name configured, using hostname 'PC-Dan'
registrar.go:75: INFO Loading registrar data from C:\logs\filebeat/.filebeat
crawler.go:77: INFO All prospectors initialised with 2 states to persist
service_windows.go:45: ERR Error: The service process could not connect to the service controller.
```
**Filebeat config file:**
```
filebeat:
  prospectors:
    -
      paths:
        - C:\logs\Frontend.trace.log

      fields:
        app: "FE"

      encoding: utf-8
      type: log
      ignore_older: 10m
      scan_frequency: 10s
    -
      paths:
        - C:\logs\Services.trace.log

      fields:
        app: "SVC"

      encoding: utf-8
      type: log
      ignore_older: 10m
      scan_frequency: 1s

  idle_timeout: 1s

############################# Output ##########################################
output:

  logstash:
    enabled: true

    hosts: ["myLogstashHost:5044"]

shipper:
```

ruflin:
As the transformer stops working on completion (EOF) it is reset every time EOF is reached.

### DIFF
--- a/harvester/log.go
+++ b/harvester/log.go
@@ -58,15 +58,14 @@ func (h *Harvester) Harvest() {
 	var line uint64 = 0 // Ask registrar about the line number
 
 	h.initOffset()
-
-	var in io.Reader = h.file
-	in = h.encoding(in)
+	in := h.encoding(h.file)
 	reader := bufio.NewReaderSize(in, h.Config.BufferSize)
 	buffer := bytes.NewBuffer(nil)
 
 	var readTimeout = 10 * time.Second
 	lastReadTime := time.Now()
 	for {
+
 		text, bytesread, err := h.readLine(reader, buffer, readTimeout)
 
 		if err != nil {
@@ -75,6 +74,9 @@ func (h *Harvester) Harvest() {
 			if err != nil {
 				return
 			} else {
+				// EOF reached, reset encoding
+				in = h.encoding(h.file)
+				reader = bufio.NewReaderSize(in, h.Config.BufferSize)
 				continue
 			}
 		}
--- a/tests/system/test_crawler.py
+++ b/tests/system/test_crawler.py
@@ -1,5 +1,8 @@
+# -*- coding: utf-8 -*-
+
 from filebeat import TestCase
 
+import codecs
 import os
 import time
 import unittest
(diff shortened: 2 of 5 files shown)


tests/system/test_prospector.py
```diff
@@ -7,7 +7,7 @@
 """
 
 
-class Test(TestCase):
+class ProspectorTest(TestCase):
```
tsg: Any reason we need this change? I find it it makes running an individual test a bit more difficult. Compare: `nostetests test_prospector.py:Test.test_ignore_old_files` with `nostetests test_prospector.py:ProspectorTest.test_ignore_old_files`.

ruflin: No, this change is not required. It felt "wrong" to have multiple classes with the same name. I can change it back.

tsg: I think it's fine to have the same name in python, where the filename is part of the namespace.

ruflin: reverted

harvester/log.go
```diff
@@ -75,6 +74,9 @@ func (h *Harvester) Harvest() {
 			if err != nil {
 				return
 			} else {
+				// EOF reached, reset encoding
```
urso: maybe extend this comment with reference to original issue, as this is a workaround for something not working as expected.

ruflin: @urso good point, I will extend the comment.

harvester/log.go
```diff
@@ -75,6 +74,9 @@ func (h *Harvester) Harvest() {
...
+				in = h.encoding(h.file)
+				reader = bufio.NewReaderSize(in, h.Config.BufferSize)
```
urso: creating a new bufio.NewReaderSize will drop the old buffers and create new one :( more garbage.
I hope even though we've seen an EOF nothing is left in the buffer (which would be unexpected).

ruflin: Should we "close" the old one more actively?

tests/system/test_crawler.py
```diff
@@ -421,6 +424,57 @@ def test_tail_on_rotate(self):
...
+            lambda: self.log_contains(
+                "Start next scan"),
+            max_timeout=15)
+
+        # Add utf-8 Chars for the first time
+        with codecs.open(testfile, "w", "utf-8") as f:
+            # Write lines before registar started
+
+            # Special encoding needed?!?
+            f.write("ニコラスRuflin".decode("utf-8") + "\n")
```
urso: why not u"..." ? e.g. on mac we still have python 2.7.10

harvester/log.go
urso: bufio.Reader has a Reset method reusing the allocated slice (but forgets all content). It has no Close method though (which is fine).

ruflin: @urso I briefly tried it with reset and the following code works:

```
in = h.encoding(h.file)
reader.Reset(in)
```

Do you prefer this option?

tests/system/test_crawler.py
ruflin: I tried that first and it didn't work. But as the above worked I didn't check in detail why u"..." broke the code.

urso: unicode in python uses 16 or 32bit presentation of character. The normal string type uses 8bit presentation. Some background on unicode frustrations in python: https://pythonhosted.org/kitchen/unicode-frustrations.html

python 3 uses unicode for string by default, but by default source code is assumed to be UTF-8: https://docs.python.org/3/howto/unicode.html

Dealing with different kind of string types and python version is quite a pain...
I think your solution not using u"..." is perfectly fine.

harvester/log.go
urso: I don't really mind, as the effects are pretty equivalent, besides creating a little less garbage.

tests/system/test_crawler.py
ruflin: @urso thanks for the insights.

harvester/log.go
ruflin: As I think the new option is cleaner and more readable I changed it here: https://github.com/ruflin/filebeat/commit/5d9b28cbe08362896c5d0064ba50d7af5c9a2730

PR #183, opened 2015-11-03
Added tests for a few encodings

tsg:
Still want to add more encodings, but the approach works very well. The test passes in 0.3 seconds :-).

### DIFF
--- a/tests/system/config/filebeat_prospectors.yml.j2
+++ b/tests/system/config/filebeat_prospectors.yml.j2
@@ -1,0 +1,20 @@
+filebeat:
+  prospectors:
+  {% for prospector in prospectors %}
+    -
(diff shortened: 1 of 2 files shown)


ruflin:
:+1: Looks good. Would it be possible to extend the test to it does the initial read and afterwards and append? Like this we make sure for all encodings the readers doesn't break.

urso:
LGTM. Would like to see the test to write the line to file, start filebeat and repeat the text like 2 times to force filebeat tailing

tsg:
I'll add testing append as well. Won't be 0.3 seconds anymore, but oh well.

tsg:
Added more encodings and I'm now also testing for an appended line.

PR #187, opened 2015-11-03
use encoding/htmlindex package for finding decoder

urso:
encoding/htmlindex package provides rather exhausitve list of known encodings
in comparison to harvester/encoding.go table.

See list of known encodings:
https://github.com/golang/text/blob/master/encoding/htmlindex/tables.go#L94


ruflin:
Should we link to http://www.w3.org/TR/encoding/ in the docs for the encoding list so we only list the most important ones?

monicasarbu:
LGTM

harvester/encoding.go
```diff
@@ -22,29 +20,16 @@ var encodings = map[string]Decoder{
 	"nop":   Plain,
 	"plain": Plain,
 
-	// utf8 (validate input)
-	"utf-8": trans(encoding.UTF8Validator),
+	// utf8 (validate input) - shadow htmlindex utf8 codecs not validating input
+	"unicode-1-1-utf-8": trans(encoding.UTF8Validator),
```
ruflin: why do we use the same utf-8 three times?

ruflin:
LGTM, needs rebase. Lets also add tests for some more encodings.

harvester/encoding.go
urso: they are the same, but with different name. These utf-8 entries shadow the utf-8 entries in htmlindex, which is equivalent to plain (no error checks)

PR #188, opened 2015-11-04
Added the fields_under_root option

tsg:
If set to true, the fields are inserted top level instead of under
the fields sub-dict. In case of conflicts, the fields overwrite existing
values.

Still need to update docs & changelog, but the code should be complete.

### DIFF
--- a/etc/filebeat.yml
+++ b/etc/filebeat.yml
@@ -36,6 +36,12 @@ filebeat:
       #  level: debug
       #  review: 1
 
+      # Set to true to store the additional fields as top level fields instead
+      # of under the "fields" sub-dictionary. In case of name conflicts with the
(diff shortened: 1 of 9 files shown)


etc/filebeat.yml
```diff
@@ -36,6 +36,10 @@ filebeat:
       #  level: debug
       #  review: 1
 
+      # Set to true to store the additional fields as top level fields instead
+      # of under the "fields" sub-dictionary.
```
ruflin: I would add a warning that this can lead to name conflicts ...

ruflin:
Code LGTM

tsg:
Commends addressed in the last commit.

monicasarbu:
LGTM

PR #211, opened 2015-11-06
Refactor log harvester to simplify it.

ruflin:
Fix issue that for some encoding bytes count instead of rune count was used for offest which lead
to restart the file reading from the beginning.

### DIFF
--- a/harvester/log.go
+++ b/harvester/log.go
@@ -13,7 +13,6 @@ import (
 	"github.com/elastic/filebeat/config"
 	"github.com/elastic/filebeat/input"
 	"github.com/elastic/libbeat/logp"
-	"unicode/utf8"
 )
 
 func NewHarvester(
@@ -78,7 +77,7 @@ func (h *Harvester) Harvest() {
 	lastReadTime := time.Now()
 
 	for {
-		text, bytesRead, err := readLine(reader, buffer, hConfig.PartialLineWaitingDuration)
+		text, err := readLine(reader, buffer, hConfig.PartialLineWaitingDuration)
 
 		if err != nil {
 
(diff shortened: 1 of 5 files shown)


harvester/log.go
```diff
@@ -337,14 +338,17 @@ func readLine(reader *bufio.Reader, buffer *bytes.Buffer, partialLineWaiting tim
...
 		// If we got a full line, return the whole line without the EOL chars (LF or CRLF)
 		if !isPartialLine {
-			// Get the str length with the EOL chars (LF or CRLF)
-			bufferSize := buffer.Len()
-			str := new(string)
-			// Remove line endings
-			*str = buffer.String()[:bufferSize-lineEndingChars(segment)]
+
+			str := buffer.String()
+			// Rune count is relevant as Size() returns also rune count
```
urso: which Size() returns rune count?

urso:
I'm feeling kinda uneasy to have offsets based on rune-count. Can we split up the harvester in some kind of input processor taking care of counting by bytes (offsets) read throught encoder + harvester using this input/file processor?

Does go have ftell like support?

harvester/log.go
ruflin: I was talking about FileInfo.Size(). Will clarify it in the docs.

ruflin:
@urso Also not too happy about this implementation because of the encoding. One option would be to do the encoding conversion after the reading so we would always read single bytes.

It seems like the Seek() could be used similar to ftell: http://stackoverflow.com/questions/10901351/fgetpos-available-in-go-want-to-find-file-position

Didn't test it yet.

ruflin:
@urso I suggest to move forward here as it seems to solve the current issue but keep iterating on top of it.

urso:
tests still failing?

tsg:
@urso we're working on it, see the discussion on chat :-)

tsg:
@ruflin, to proceed here, should we take the harvester refactoring and leave the offset fix for a different PR?

ruflin:
Sounds good to me. The question is which one of the broken offset implementation should we take? The main problem with the initial on is that it starts to read the file from the beginning again quite often.

ruflin:
This change will heavily conflict with https://github.com/elastic/filebeat/pull/224. Some changes will not be even required anymore. I suggest to first get https://github.com/elastic/filebeat/pull/224 in and then reapply the changes that are needed. This probably then can even be combined with https://github.com/elastic/filebeat/pull/222 

@urso FYI

ruflin:
Closing in favor of https://github.com/elastic/filebeat/pull/232

PR #218, opened 2015-11-10
Add auto update for config file

ruflin:
This should simplify the release process and make sure config file is always up-to-date.


ruflin:
https://github.com/elastic/libbeat/pull/284 should be merged first and then the update should be run again including Godeps enabled.

Makefile
```diff
@@ -84,3 +84,15 @@ install-cfg:
...
+
+# Update expects the most recent version of libbeat in the GOPATH
+# Updates libbeat and config file
+.PHONY: update
+update:
+	# Update Libbeat
+	godep update github.com/elastic/libbeat/...
+	# Update config
+	#rm etc/filebeat.yml
+	echo "### THIS FILE IS GENERATED. DO NOT EDIT ###" > etc/filebeat.yml
```
tsg: If we put this, we should remove it before packaging, because we don't want that line in the installed packages :-).

Makefile
```diff
@@ -84,3 +84,15 @@ install-cfg:
...
+	cat etc/beat.yml ${GOPATH}/src/github.com/elastic/libbeat/etc/libbeat.yml >> etc/filebeat.yml
```
tsg: I thought godeps don't copy the yml files?

ruflin: I doesn't. That is why the requirement of update is that libbeat is in the gopath (see comment above).

Makefile
ruflin: Shouldn't we remove the complete makefile before packaging?

tsg: I meant the line from `etc/filebeat.yml`, not the Makefile line.

Makefile
tsg: I see, ok.

Makefile
ruflin: Good point. Also the beat.yml file.

ruflin:
Related to https://github.com/elastic/libbeat/issues/280

Makefile
ruflin: I added an issue so we don't forget about it: https://github.com/elastic/beats-packer/issues/22

tsg: To be honest, I don't see that much benefit in having the line to complicate the build process. If you want to have it, IMHO best place to put it is in the `install-cfg` line from the beat Makefile, rather than in beats-packer.

etc/filebeat.yml
```diff
@@ -215,19 +230,24 @@ output:
...
-
   ### Logstash as output
+
   #logstash:
     # Uncomment out this option if you want to output to Logstash. The default is false.
     #enabled: true
 
     # The Logstash hosts
-    #hosts: ["localhost:5044"]
+    #hosts: ["localhost:5044", "logstash:5045"]
```
tsg: Hmm, how did this came back?

scripts/Makefile
```diff
@@ -0,0 +1,177 @@
+#/bin/bash
```
tsg: Is moving some of the Makefile under `scripts/` and intermediary step to getting it from libbeat?  Or is there another motivation?

Makefile
```diff
@@ -1,73 +1,13 @@
...
 
-filebeat: $(GOFILES)
-	# first make sure we have godep
-	go get github.com/tools/godep
-	$(GODEP) go build
 
+# Runs all tests and generates the coverage reports
+.PHONY: integration-tests-environment
+integration-tests-environment:
+	echo "No integration test Enviornment for filebeat"
```
tsg: `make test` fails for me, with this:

<pre>
make test
Makefile:9: warning: overriding commands for target `integration-tests-environment'
scripts/Makefile:96: warning: ignoring old comma
</pre>

etc/filebeat.yml
```diff
@@ -153,11 +157,14 @@ output:
     # IPv6 addresses should always be defined as: https://[2001:db8::1]:9200
     hosts: ["localhost:9200"]
 
-    # Optional protocol and basic auth credentials. These are deprecated.
+    # Optional protocol and basic auth credentials. These are deprectad
```
tsg: This change was done by mistake, I assume.

tsg:
Generally LGTM, please see comments.

etc/filebeat.yml
ruflin: This is coming from the libbeat config. Will change it there. https://github.com/elastic/libbeat/blob/master/etc/libbeat.yml#L105

scripts/Makefile
ruflin: Yes. The Makefile part under scripts is the one that is common between all beats and is coming from libbeat. `sed` is used to replace some parts which cannot be configured with variables. Would you prefer to have the file in a different location?

Makefile
ruflin: The first error is expected as I overwrite it actually  as it is empty for filebeat. But the better solution is to wrap it into the if statement so no error is thrown. I have to check what the second error is.

etc/filebeat.yml
ruflin: This is coming from libbeat. I fixed it there. That is the nice part about the the script. We will have consistency across all beats.

ruflin:
Failing of flaky test does not seem to be related to this PR.

urso:
LGTM

PR #222, opened 2015-11-11
Fixing flaky test by depending on output instead of log lines

ruflin:


ruflin:
This does not work yet as expected.

tests/system/test_crawler.py
```diff
@@ -253,8 +253,7 @@ def test_file_disappear_appear(self):
 
         # Let it read the file
         self.wait_until(
-            lambda: self.log_contains(
-                "Registrar: processing 6 events"),
+            lambda: self.output_has(11),
```
tsg: Just a tiny side note: I usually prefer the `self.output_has(lines=11)` because it's a bit more readable and the intention was that in the future we'd have more abstractions, like `self.output_has(keyword="hello")`. Might have beed a bad idea :-).

ruflin: Will adjust it.

Unfortunately the test is somehow still broken ...

ruflin:
The changes I have to make here will potentially also conflict with https://github.com/elastic/filebeat/pull/224 We should first push https://github.com/elastic/filebeat/pull/224 in.

ruflin:
Closing as fix covered by https://github.com/elastic/filebeat/pull/232

PR #224, opened 2015-11-12
extract line reader handling decoding from harvester

urso:
split line reader from harvester. reader separates lines, applies the configured codec + while reading lines it keeps track of total bytes consumed from raw input stream (speak: correct file offset in bytes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/elastic/libbeat/compare/1.0.0-rc1...HEAD)
 
 ### Backward Compatibility Breaks
-- Removed utf-16be-bom encoding support
+- Removed utf-16be-bom encoding support. Support will be added with fix for #205
 
 ### Bugfixes
 - Filebeat will now exit if a configuration error is detected. #198
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,13 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/elastic/libbeat/compare/1.0.0-rc1...HEAD)
 
 ### Backward Compatibility Breaks
+- Removed utf-16be-bom encoding support
 
 ### Bugfixes
 - Filebeat will now exit if a configuration error is detected. #198
 - Fix to enable prospector to harvest existing files that are modified. #199
+- Improve line reading and encoding to better keep track of file offsets based
+  on encoding. #224
 
 ### Added
 
--- a/harvester/encoding.go
+++ b/harvester/encoding.go
@@ -8,6 +8,7 @@ import (
 	"golang.org/x/text/encoding/charmap"
 	"golang.org/x/text/encoding/htmlindex"
 	"golang.org/x/text/encoding/simplifiedchinese"
+	"golang.org/x/text/transform"
 )
 
 // Decoder wraps a reader for decoding input to utf-8 on read.
@@ -18,12 +19,10 @@ var encodings = map[string]encoding.Encoding{
 	"nop":   Plain,
 	"plain": Plain,
 
-	/*
-		// utf8 (validate input) - shadow htmlindex utf8 codecs not validating input
-		"unicode-1-1-utf-8": trans(encoding.UTF8Validator),
-		"utf-8":             trans(encoding.UTF8Validator),
-		"utf8":              trans(encoding.UTF8Validator),
-	*/
+	// utf8 (validate input) - shadow htmlindex utf8 codecs not validating input
+	"unicode-1-1-utf-8": utf8Encoding{},
+	"utf-8":             utf8Encoding{},
+	"utf8":              utf8Encoding{},
 
 	// utf16
 	// "utf-16be-bom": unicode.UTF16(unicode.BigEndian, unicode.UseBOM),
@@ -41,6 +40,20 @@ var encodings = map[string]encoding.Encoding{
 // Plain file encoding not transforming any read bytes.
 var Plain = encoding.Nop
 
+// UTF-8 encoding copying input to output sequence replacing invalid UTF-8
+// converted to '\uFFFD'.
+//
+// See: http://encoding.spec.whatwg.org/#replacement
+type utf8Encoding struct{}
+
+func (utf8Encoding) NewDecoder() transform.Transformer {
+	return encoding.Replacement.NewEncoder()
+}
+
+func (utf8Encoding) NewEncoder() transform.Transformer {
+	return encoding.Replacement.NewEncoder()
+}
+
 // Find returns
 func findEncoding(name string) (encoding.Encoding, bool) {
 	if name == "" {
@@ -54,17 +67,3 @@ func findEncoding(name string) (encoding.Encoding, bool) {
 	codec, err := htmlindex.Get(name)
 	return codec, err == nil
 }
-
-/*
-func nopEnc(r io.Reader) io.Reader { return r }
-
-func enc(encoding encoding.Encoding) Decoder {
-	return trans(encoding.NewDecoder())
-}
-
-func trans(t transform.Transformer) Decoder {
-	return func(r io.Reader) io.Reader {
-		return transform.NewReader(r, t)
-	}
-}
-*/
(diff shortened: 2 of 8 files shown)


harvester/encoding.go
```diff
@@ -8,41 +8,41 @@ import (
...
-	// utf8 (validate input) - shadow htmlindex utf8 codecs not validating input
-	"unicode-1-1-utf-8": trans(encoding.UTF8Validator),
-	"utf-8":             trans(encoding.UTF8Validator),
-	"utf8":              trans(encoding.UTF8Validator),
+	/*
+		// utf8 (validate input) - shadow htmlindex utf8 codecs not validating input
+		"unicode-1-1-utf-8": trans(encoding.UTF8Validator),
+		"utf-8":             trans(encoding.UTF8Validator),
+		"utf8":              trans(encoding.UTF8Validator),
+	*/
```
urso: TODO: create full encoding based on utf8 validator.

harvester/encoding.go
```diff
@@ -8,41 +8,41 @@ import (
...
 
 	// utf16
-	"utf-16be-bom": enc(unicode.UTF16(unicode.BigEndian, unicode.UseBOM)),
+	// "utf-16be-bom": unicode.UTF16(unicode.BigEndian, unicode.UseBOM),
```
urso: decided to disable "utf-16be-bom" in this PR, as handling this codec needs some more logic on start. To be done in another PR.

harvester/log.go
```diff
@@ -292,55 +294,37 @@ func lineEndingChars(line []byte) int {
...
-			str := new(string)
-			// Remove line endings
-			*str = buffer.String()[:bufferSize-lineEndingChars(segment)]
-			// Reset the buffer for the next line
-			buffer.Reset()
-			return str, bufferSize, nil
+		// test for no file updates longer than partialLineWaiting
+		if time.Since(*lastReadTime) >= partialLineWaiting {
+			// return partial line:
+			// XXX: we rather want to drop partial line (like orignial implementation?)
```
urso: opinions?

harvester/log.go
```diff
@@ -65,10 +63,16 @@ func (h *Harvester) Harvest() {
 	// Load last offset from registrar
 	h.initOffset()
 
-	in := h.encoding(h.file)
+	// TODO: newLineReader uses additional buffering to deal with encoding and testing
+	//       for new lines in input stream. Simple 8-bit based encodings, or plain
+	//       don't require 'complicated' logic.
```
urso: TODO: but can be done in another PR. Current solution maybe not optimal for performance + memory usage, but not incorrect.

urso:
requires elastic/libbeat#290

harvester/reader_test.go
```diff
@@ -0,0 +1,86 @@
...
+			continue
+		}
+		for i := range test.lines {
+			expected := test.lines[i]
+			actual := readLines[i]
+			assert.Equal(t, expected, actual)
+			assert.Equal(t, expectedCount[i], byteCounts[i])
+		}
+	}
+}
```
urso: TODO: add tests for reading partial lines (missing EOL)

harvester/log.go
ruflin: I would actually return partial lines. I think it is a rare case and it is better to have a line twice then not at all.

harvester/log.go
```diff
@@ -292,55 +294,37 @@ func lineEndingChars(line []byte) int {
...
-func readLine(reader *bufio.Reader, buffer *bytes.Buffer, partialLineWaiting time.Duration) (*string, int, error) {
-
-	lastSegementTime := time.Now()
-	isPartialLine := true
-
+func readLine(
+	reader *lineReader,
+	lastReadTime *time.Time,
+	partialLineWaiting time.Duration,
+) (string, int, error) {
```
ruflin: we had *string before as it made it possible to differentiate between an empty line ("") and no line (nil). At the same time error should be set in case it is not a line.

harvester/reader.go
```diff
@@ -0,0 +1,215 @@
...
+
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/transform"
+
+	"github.com/elastic/libbeat/common/streambuf"
+)
+
+// timedReader keeps track of last time bytes have been read from underlying
+// reader.
+type timedReader struct {
```
ruflin: Instead of timedReader I would probably call it lineReader or something similar. My first impression with the name was that it is a reader that reads and a predefined timing.

harvester/reader.go
```diff
@@ -0,0 +1,215 @@
...
+
+	nl        []byte
+	inBuffer  *streambuf.Buffer
+	outBuffer *streambuf.Buffer
+	inOffset  int // input buffer read offset
+	byteCount int // number of bytes decoded from input buffer into output buffer
+	decoder   transform.Transformer
+}
+
+const maxConsecutiveEmptyReads = 100
```
ruflin: Is this value based on some assumptions you made? Or related to the BufferSize?

harvester/reader_test.go
ruflin: @urso There are some system tests for this already.

ruflin:
I like this approach a lot as it simplifies the harvester, as in log.go it doesn't have to be dealt with buffering and encoding. I'm thinking if we should move reader.go in the long term to libbeat or even contribute parts to the Go source code.

harvester/log.go
urso: before this PR, partial lines have been dropped, but input stream has been consumed.

This PR returns the partial line, but bytes are also consumes the bytes returned. That is the substring returned here will not be repeated in subsequent calls. I was not sure about wanted semantics here. Maybe we change this to repeat the complete line if any bytes have been read in meantime.

harvester/reader_test.go
urso: cool. As this encoding stuff is quite fragile we should check the combined coverage report and see if we're missing something.

harvester/log.go
urso: the harvester worker is not using the returned line if error!=nil. And readLine reads until EOF or some bytes have been consumed.

harvester/reader.go
urso: Don't like timedReader either, but lineReader is no good name too.

this reader sits between the file input and the line reader. It's only purpose is to update an internal timestamp whenever any bytes have been read from file. The timestamp is used in readLine with partialLineWaiting to have some kind of 'keep-alive' ;)

harvester/reader.go
urso: I did follow source code of most functions used before PR in order to be as close in behavior to original implementation as possible. This variable + loop on read is 'inspired' by standard library.

urso:
Not sure about moving to libbeat as is. Behavior is tailored for filebeat harvester (but quite generic, true). When moving we should improve on the interfaces in reader.go (naming is really bad).

harvester/log.go
ruflin: If we drop the partial line, we definitively should retry later at the beginning of the line. What is easier to implement?

harvester/reader.go
ruflin: Not "urgent" at the moment. We can also rename it later when we have a better name.

harvester/reader.go
ruflin: :+1:

harvester/log.go
urso: easiest is as is. When restarting at beginning of line we have to check line was updated since last partial return, else we might repeat lines quite a load of times.

ruflin: Ok, lets keep it for the moment like this.

harvester/log.go
ruflin: I think this would mainly matter for really large lines.

CHANGELOG.md
```diff
@@ -6,10 +6,13 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/elastic/libbeat/compare/1.0.0-rc1...HEAD)
 
 ### Backward Compatibility Breaks
+- Removed utf-16be-bom encoding support
```
andrewkroh: From a user perspective, it would be nice to answer the "why" question or link to an issue.

urso: I will re-add utf-16-bom soonish.

andrewkroh:
Looks good.

PR #225, opened 2015-11-12
Update appveyor configuration to publish cross-compiled artifacts.

andrewkroh:
* Removed hipchat notifications from appveyor.yml so that developers setup CI jobs for their forks. (Need to re-add them through the AppVeyor UI.)
* Renamed .appveyor.yml to appveyor.yml which is the default AppVeyor filename.

Here is a working build: https://ci.appveyor.com/project/andrewkroh/filebeat/build/8/artifacts

I might want to setup a build matrix on AppVeyor so that cross-compilation only happens when building the master branch.


ruflin:
@andrewkroh We renamed appveyor.yml to .appveyor.yml so it is a hidden file like .travis.yml. I would prefer it that way even though I see that it means +1 step for the setup. @tsg Your toughts?

appveyor.yml
```diff
@@ -0,0 +1,197 @@
...
+# After testing, cross-compile and publish artifacts.
+after_test:
+  - ps: cd c:\gopath\src\github.com\elastic\filebeat
+  - appveyor AddCompilationMessage "Starting cross-compile"
+  - set GOOS=linux
+  - set GOARCH=386
+  - set GOARM=
+  - set GO386=
+  - go build -o bin/filebeat-linux-386
+  - ps: Push-AppveyorArtifact bin/filebeat-linux-386
```
ruflin: After https://github.com/elastic/filebeat/pull/218 all will be in the build folder. Will it make sense to just push the full folder or is it preferrable to have every single file downloadable?

andrewkroh: By pushing each artifact, you can see the build progression if you are watching the artifacts tab.

ruflin:
LGTM

@andrewkroh The most recent commit seems to fail?

andrewkroh:
I am still working on getting the test output folder archived. Getting this error at the moment: `tests\system\run:  WARNING: The system cannot find the file specified`

andrewkroh:
@ruflin I had the wrong path. It is working as expected (though there is still a test failure.) Here are the archived outputs from the tests: https://ci.appveyor.com/project/elastic-beats/filebeat/build/397/artifacts

ruflin:
That is very nice with the artifacts. The test that is failing is not related to your build.

PR #233, opened 2015-11-14
File state offset based on input bytes

urso:
Update file state input to be based on bytes read from input to properly handle
offsets for different newline patterns and encodings.


ruflin:
@urso This actually broke on of the tests on windows. Still have to check if the test is wrong or the change: https://ci.appveyor.com/project/elastic-beats/filebeat/build/420#L78

Issue #205, opened 2015-11-05
Encoding utf-16be-bom doesn't work


tsg:
Here are steps to reproduce (I did this on Windows, but I think it's not platform specific):

* Configure "utf-16be-bom" in filebeat.yml
* Start filebeat
* Create a log line to a new file: `"hello1" >> test.log`
* This works just fine:

<pre>
2015/11/05 02:58:28.105090 publish.go:98: DBG  Publish: {
  "@timestamp": "2015-11-05T02:58:20.641Z",
  "count": 1,
  "fields": null,
  "fileinfo": {},
  "input_type": "log",
  "line": 1,
  "message": "hello1",
  "offset": 0,
  "shipper": "WIN-UMHB36DFNOC",
  "source": "c:\\programdata\\logs\\test.log",
  "type": "log"
}
</pre>

* Append another log line: `"hello2" >> test.log`
* Filebeat shows the following errors:

<pre>
2015/11/05 02:59:51.041687 log.go:165: DBG  harvest: "c:\\programdata\\logs\\test.log" position:8 (offset snapshot:8)
2015/11/05 02:59:57.048521 log.go:257: ERR Unexpected state reading from c:\programdata\logs\test.log; error: transform:
 short source buffer
2015/11/05 02:59:57.049465 log.go:126: ERR File reading error. Stopping harvester. Error: transform: short source buffer
</pre>

I suspect we have a problem with encodings that use a "byte order mark" at the beginning of the file, because we reset the encoder when it hits EOF, and then the decoder expects again a BOM but that never comes.

PR #235, opened 2015-11-15
Introduce encoding factory in open

urso:
Introduce encoding factory to determine encoding from file on open and provide
utf16 encodings with BOM support checking first 2 bytes from file.


urso:
This PR makes the utf16 bom based encodings available again. See #205.

Limitations:
- files requires at least 2 bytes to be written. Otherwise open fails and harvester will not start. 
- BOM support requires seekable source => no love for stdin.

harvester/log.go
```diff
@@ -230,12 +230,32 @@ func (h *Harvester) open() error {
...
+
+	encoding, err := h.encoding(h.file)
+	if err != nil {
+		h.file.Close()
+		return nil, err
 	}
 
-	h.setFileOffset()
+	// update file offset
+	if h.Offset > 0 {
```
ruflin: I would still keep this inside the setFileOffset method

ruflin:
@urso It seems to break the appveyor tests ...

ruflin:
@urso about the stdin issue. Stdin should NOT be part of the log harvester. I'm aware that it still inside but we should extract it later as an independent type so the above will not be an issue.

urso:
I see this failing badly with stdin in general. Needs some more thinking

harvester/encoding/utf16.go
```diff
@@ -0,0 +1,106 @@
...
+	// read Byte Order Marker (BOM)
+	var buf [2]byte
+	n, err := in.Read(buf[:])
+	if err != nil {
+		in.Seek(offset, os.SEEK_SET)
+		return nil, err
+	}
+	if n < 2 {
+		in.Seek(offset, os.SEEK_SET)
+		return nil, transform.ErrShortSrc
```
ruflin: Lets add a Info message here.

harvester/encoding/utf16.go
```diff
@@ -0,0 +1,106 @@
...
+	switch {
+	case buf[0] == 0xfe && buf[1] == 0xff:
+		inEndiannes = bigEndian
+	case buf[0] == 0xff && buf[1] == 0xfe:
+		inEndiannes = littleEndian
+	}
+
+	// restore offset if BOM is missing or this function was not
+	// called with read pointer at beginning of file
+	if !keepOffset || inEndiannes == unknownEndianess {
```
ruflin: This basically mean, it is probably not utf16, right?

harvester/encoding/utf16.go
```diff
@@ -0,0 +1,106 @@
...
+		in, ok := in_.(io.ReadSeeker)
+		if !ok {
+			return nil, ErrUnsupportedSourceTypeBOM
+		}
+
+		return utf16Seekable(in, e)
+	}
+}
+
+func utf16Seekable(in io.ReadSeeker, endianness endianness) (Encoding, error) {
```
ruflin: Would it solve your problem, if filebeat wouldn't ever start reading empty files?

harvester/log.go
```diff
@@ -152,67 +155,88 @@ func (h *Harvester) backOff() {
...
-// Sets the offset of the file to the right place. Takes configuration options into account
-func (h *Harvester) setFileOffset() {
-	if h.Offset > 0 {
-		h.file.Seek(h.Offset, os.SEEK_SET)
-	} else if h.Config.TailFiles {
-		h.file.Seek(0, os.SEEK_END)
-	} else {
-		h.file.Seek(0, os.SEEK_SET)
-	}
+func (h *Harvester) openStdin() (encoding.Encoding, error) {
```
ruflin: :+1:

ruflin:
In general looks good to me. To be honest it is very hard to figure out if the all the checks are still made in the identical order as they were made before as quite a few parts were moved around. But as our tests cover most of the edge cases, I'm quite confident that this will work as expected. Good for me to move forward.

harvester/encoding/utf16.go
urso: this would help. On the other hand check opening loop. In case of file being empty, the loop will retry after backing off. For now it's an okish solution I would say.

harvester/encoding/utf16.go
urso: No, problem is the BOM is optional. Missing BOM does not say anything about the encoding. If missing the codec must fallback to either big endian or little endian. Default fallback should be big endian, but windows uses little endian by default.

harvester/encoding/utf16.go
urso: +1

urso: added info to open function

harvester/encoding/utf16.go
ruflin: As it seems like we can't guarantee to detect it correctly, should we make it configurable?

urso: make what configurable? It already is. There are 3 codecs: utf-16be-bom (fallback to BigEndian), utf-16le-bom (fallback to LittleEndian) and utf-16-bom (fail if BOM is missing).

PR #247, opened 2015-11-17
Input type fix

ruflin:
* Removing line from document which is sent. This was already mentioned in RC1 that it should be removed. Was not removed in all places.
* Removing fileinfo from event. This was always empty.


docs/index.asciidoc
```diff
@@ -1,6 +1,6 @@
 = Filebeat
-:libbeat: http://www.elastic.co/guide/en/beats/libbeat/1.0.0-rc2
-:version: 1.0.0-rc2
+:libbeat: http://www.elastic.co/guide/en/beats/libbeat/master
+:version: master
```
tsg: Can you rebase to remove these changes? They would break the docs.

PR #254, opened 2015-11-17
Use type: dict for fields of that type

tsg:
It used to be type: group, but that is interpreted by the docs
generating script to mean that it has documented sub-fields.

@dedemorton can you review the result, please?

### DIFF
--- a/docs/fields.asciidoc
+++ b/docs/fields.asciidoc
@@ -90,12 +90,20 @@ required: True
 Content of line read from log file given by source.
 
 
-==== fields Fields
+==== fields
+
+type: dict
+
+required: False
 
 Contains user configurable fields.
 
 
-==== fileinfo Fields
+==== fileinfo
+
+type: dict
+
+required: False
 
 Operating System specific file information use to identify the source file. For example on linux an inode might be reported.
 
--- a/docs/fields.asciidoc
+++ b/docs/fields.asciidoc
@@ -99,12 +99,3 @@ required: False
 Contains user configurable fields.
 
 
-==== fileinfo
-
-type: dict
-
-required: False
-
-Operating System specific file information use to identify the source file. For example on linux an inode might be reported.
-
-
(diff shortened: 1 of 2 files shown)


docs/fields.asciidoc
```diff
@@ -90,12 +90,20 @@ required: True
...
+
+type: dict
+
+required: False
 
 Contains user configurable fields.
 
 
-==== fileinfo Fields
+==== fileinfo
```
ruflin: fileinfo was removed, so this part can be removed I think

etc/fields.yml
```diff
@@ -66,13 +66,13 @@ log:
         Content of line read from log file given by source.
 
     - name: fields
```
ruflin: See above

etc/fields.yml
```diff
@@ -66,13 +66,13 @@ log:
...
-      type: group
+      type: dict
       required: false
       description: >
         Contains user configurable fields.
 
     - name: fileinfo
-      type: group
+      type: dict
```
ruflin: As this has to be built by https://github.com/elastic/filebeat/blob/master/scripts/generate_field_docs.py I'm not sure this supports dict yet? I could find group but not dict.

monicasarbu: Yes, `dict` is not supported by the script. But you said earlier that `fileinfo` was removed. If so, then we don't need to generate a `dict` field.

ruflin: Dict is also used for the fields above.

tsg: I tried and the output seemed fine. "dict" is used in Packetbeat where the system tests check that all fields are documented, that's why I chose it.

ruflin: Ok, it seems like dict is properly generated by the else: https://github.com/elastic/filebeat/blob/master/scripts/generate_field_docs.py#L41

ruflin:
LGTM

etc/fields.yml
monicasarbu: Actually the type that you set there will be added as it is to the doc. At first I thought `dict` is a special case, but it's not. Only `group` is different.

PR #257, opened 2015-11-18
Fix State issue on discovering of new files

ruflin:
Not listening on the state channel properly anymore, it could lead to the harvestor getting stuck.


crawler/registrar.go
```diff
@@ -119,6 +122,14 @@ func (r *Registrar) Run() {
 	}
 }
 
+// Listents for new file states to be persisted.
+func (r *Registrar) stateListener() {
+	for event := range r.Persist {
+		r.State[*event.Source] = event
```
urso: this is asking for trouble? Registrar.Run main loop also updates this map. (Not sure if tested with race detector, but I think race detector will only mark races if they happen during tests... quite limited)

urso: what happens if 'fetch' methods update the state using Persist and state for same file is overwritten same moment from main loop?

ruflin: @urso agree. see my new version :-)

crawler/registrar.go
```diff
@@ -87,15 +87,25 @@ func (r *Registrar) Run() {
 	// Writes registry on shutdown
 	defer r.writeRegistry()
 
+	var state *FileState = nil
+
```
urso: keep state variable more local to where it's used. if declared inside for, no need to set to nil again

crawler/registrar.go
andrewkroh: You can also run the application with the race detector on which can be useful for ad hoc testing.

Something like this is what I use (not sure if there is a better way of specifying the .go files):
`go run -race $(ls *.go | grep -v _test) -c etc/filebeat.dev.yml -e -v`

crawler/registrar.go
```diff
@@ -87,15 +87,25 @@ func (r *Registrar) Run() {
...
 		select {
 		case <-r.done:
 			logp.Info("Ending Registrar")
 			return
+		// Treats new log files to persist with higher priority then new events
+		case state = <-r.Persist:
 		case events = <-r.Channel:
 		}
 
+		if state != nil {
```
urso: no need for if by moving the body into select statement (it's just 2 lines) then the select becomes:

    case state := <- r.Persist:
        r.State[*state.Source] = state
        logp.Debug(...)

so state variable does not pollute your namespace. Rest of the body can go into

case events := <-r.Channel:
   ... (maybe calling another function handling the multiple events for having cleaner event loop)

beat/spooler.go
```diff
@@ -88,13 +88,13 @@ func (s *Spooler) Run() {
 
 			// Spooler is full -> flush
 			if len(s.spool) == cap(s.spool) {
-				logp.Debug("spooler", "Flushing spooler because spooler full. Events flushed: %s", len(s.spool))
+				logp.Debug("spooler", "Flushing spooler because spooler full. Events flushed: %v", len(s.spool))
 				s.flush()
 			}
 		case <-ticker.C:
 			// Flush periodically
 			if now := time.Now(); now.After(s.nextFlushTime) {
```
andrewkroh: Not new (nor a bug), but I noticed that `now` isn't needed in that scope so it could simple be `if time.Now().After(s.nextFlushTime)`.

crawler/prospector.go
```diff
@@ -285,6 +286,7 @@ func (p *Prospector) checkNewFile(newinfo *ProspectorFileStat, file string, outp
 	if newinfo.Fileinfo.ModTime().Before(p.lastscan) &&
 		time.Since(newinfo.Fileinfo.ModTime()) > p.ProspectorConfig.IgnoreOlderDuration {
 
+		logp.Debug("prospector", "Fetching old state of file to resume")
```
andrewkroh: I think it would be helpful to reference the file name in the message.

urso:
Is init loop in crawler still required: https://github.com/elastic/filebeat/pull/257/files#diff-8cbd024a06b941eb89f24b91f72757dcR60

all in all fix ok for now. But I think registrar should be one of the first subsystems being started.

crawler/prospector.go
```diff
@@ -386,8 +388,7 @@ func (p *Prospector) Stop() {
 // Check if the given file was renamed. If file is known but with different path,
```
andrewkroh: This documentation seem to be out of date.

crawler/prospector.go
ruflin: Fixed

ruflin:
@urso it is still required as it must wait for all prospectors to be setup. We should do a refactoring here after GA.

crawler/prospector.go
ruflin: Updated

beat/spooler.go
ruflin: Fixed.

crawler/registrar.go
```diff
@@ -88,34 +88,40 @@ func (r *Registrar) Run() {
...
 		}
 
-		logp.Debug("registrar", "Processing %d events", len(events))
-
-		// Take the last event found for each file source
-		for _, event := range events {
-			if !r.running {
-				break
-			}
+		if e := r.writeRegistry(); e != nil {
```
urso: hmm... do we want to call writeRegistry even on messages from r.Persist? This will flush the complete known state. Is this in line with recent behavior?

crawler/prospector.go
ruflin: Fixed

crawler/registrar.go
ruflin: I think it is a good idea to write the registar also when new files pop up.

crawler/prospector.go
```diff
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"fmt"
```
urso: what happened here?

ruflin: IntelliJ :-) Fixed

urso:
LGTM

urso:
@ruflin ready to merge. Maybe squash some commits first.

ruflin:
Closing in favor of https://github.com/elastic/filebeat/pull/261

PR #265, opened 2015-11-19
Ensure that *_p fields are cast to a float - otherwise if the first d…

robin13:
…ocument contains a '0' in a percentage field it will be cast as a long which will make kibana graphs very ... flat.


monicasarbu:
@robin13 Thanks for the improvement.

ruflin:
This belongs to topbeat and not filebeat. @robin13 will reopen it there.

PR #266, opened 2015-11-19
Improvements for the migration guide

tsg:
Added a section about the changes to the output fields and a few other
clarifications/fixes.

### DIFF
--- a/docs/migration.asciidoc
+++ b/docs/migration.asciidoc
@@ -12,13 +12,13 @@ Filebeat introduces the following major changes:
   changed.
 * Command line options were removed and moved to the configuration file.
 * Configuration options for outputs are now inherited from libbeat. For details, see the {libbeat}/index.html[Beats Platform Reference].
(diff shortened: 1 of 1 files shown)


andrewkroh:
LGTM.

docs/migration.asciidoc
```diff
@@ -90,7 +90,10 @@ with this configuration in Logstash Forwarder:
       ],
 
       # A dictionary of fields to annotate on each event.
-      "fields": { "type": "syslog" }
+      "fields": {
+        "type": "syslog",
+        "service": "example-service"
```
ruflin: I like that you extended this. I was thinking about adding even more fields here to make it clearly differentiate from document_type and input_type. Additional option could be: "zone": "us-east-1". I would also replace "example-service" through apache to make it more specific.

ruflin:
LGTM. I don't like that Appveyor is flaky again. Will look into this.

monicasarbu:
+1

docs/migration.asciidoc
monicasarbu: I agree "apache" is better than "example-service". It's just easier to understand.

PR #267, opened 2015-11-20
Introduce prospector test for stdin

ruflin:


urso:
+1 let's add some stdin test with different encodings

PR #268, opened 2015-11-20
Fix flaky disappear-appear test

ruflin:
The test was failing because sometimes filebeat was reading 2 times 3 lines instead of one times 6 times.
This is caused by the params set to a low frequency. As it only matters that all lines are read in the end
the test was adapted to check for the total line number instead of log messages which is more reliable


tsg:
LGTM, waiting for travis.

PR #280, opened 2015-11-23
prospector and harvester state sharing refactoring

urso:
Refactor sharing processing state between prospector and harvester
by moving crawler.ProspectorFileStat to harvester.FileStat
and sharing the FileStat between harvester and prospector instead
of just the channel.

Helps with problems of the prospector reassigning the channel besides
having the harvester already initialized with another channel. So harvester
will send processing return value to correct channel.


crawler/prospector.go
```diff
@@ -368,13 +357,13 @@ func (p *Prospector) checkExistingFile(newinfo *ProspectorFileStat, newFile *inp
 		// We only need to keep it for the remainder of this iteration then we can assume it was deleted and forget about it
 		p.missingFiles[file] = oldFile.FileInfo
 
-	} else if len(newinfo.Harvester) != 0 && oldFile.FileInfo.ModTime() != newinfo.Fileinfo.ModTime() {
+	} else if newinfo.Finished() && oldFile.FileInfo.ModTime() != newinfo.Fileinfo.ModTime() {
```
urso: this Finished() check requires 'Continue(lastinfo)' being called before

urso:
closed in favor of new PR against master

PR #284, opened 2015-11-23
prospector and harvester state sharing refactoring

urso:
Refactor sharing processing state between prospector and harvester
by moving crawler.ProspectorFileStat to harvester.FileStat
and sharing the FileStat between harvester and prospector instead
of just the channel.

Helps with problems of the prospector reassigning the channel besides
having the harvester already initialized with another channel. So harvester
will send processing return value to correct channel.

### DIFF
--- a/crawler/prospector.go
+++ b/crawler/prospector.go
@@ -14,19 +14,12 @@ import (
 
 type Prospector struct {
 	ProspectorConfig cfg.ProspectorConfig
-	prospectorList   map[string]ProspectorFileStat
+	prospectorList   map[string]harvester.FileStat
 	iteration        uint32
 	lastscan         time.Time
 	registrar        *Registrar
 	missingFiles     map[string]os.FileInfo
 	running          bool
-}
-
-// Contains statistic about file when it was last seend by the prospector
-type ProspectorFileStat struct {
-	Fileinfo      os.FileInfo /* the file info */
-	Harvester     chan int64  /* the harvester will send an event with its offset when it closes */
-	LastIteration uint32      /* int number of the last iterations in which we saw this file */
 }
 
 // Init sets up default config for prospector
@@ -62,7 +55,7 @@ func (p *Prospector) setupProspectorConfig() error {
 	}
 
 	// Init File Stat list
-	p.prospectorList = make(map[string]ProspectorFileStat)
+	p.prospectorList = make(map[string]harvester.FileStat)
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


crawler/prospector.go
```diff
@@ -14,7 +14,7 @@ import (
 
 type Prospector struct {
 	ProspectorConfig cfg.ProspectorConfig
-	prospectorList   map[string]ProspectorFileStat
+	prospectorList   map[string]harvester.FileStat
```
ruflin: Make sense. Prospector should know about harvester but not the other way around.

crawler/prospector.go
```diff
@@ -187,7 +180,7 @@ func (p *Prospector) Run(spoolChan chan *input.FileEvent) {
 
 		// Clear out files that disappeared and we've stopped harvesting
 		for file, lastinfo := range p.prospectorList {
-			if len(lastinfo.Harvester) != 0 && lastinfo.LastIteration < p.iteration {
+			if lastinfo.Finished() && lastinfo.LastIteration < p.iteration {
```
ruflin: :+1: here. Later we should probably rename lastinfo to something better and more describtive. Same for newinfo

harvester/harvester.go
```diff
@@ -15,13 +16,20 @@ type Harvester struct {
...
 	SpoolerChan      chan *input.FileEvent
 	encoding         encoding.Encoding
 	file             *os.File /* the file being watched */
 	backoff          time.Duration
 }
 
+// Contains statistic about file when it was last seend by the prospector
+type FileStat struct {
+	Fileinfo      os.FileInfo /* the file info */
+	Return        chan int64  /* the harvester will send an event with its offset when it closes */
```
ruflin: Not sure about the naming of Return here, but still better then Harvester.

ruflin: Isn't this just the CurrentOffset?

ruflin:
This changes simplifies the handling of the harvester. Further refactoring (potential renaming) should happen in a next step.

harvester/harvester.go
urso: The harvester is run in parallel to the prospector (in fact multiple harvesters are run in parallel). Once the harvester finishes, it returns the last offset to 'Return' (I thought of the channel as asynchronous return value, like a future or promise). The channel is used to 1.) detect the harvester having finished and 2.) get the last offset for the prospector to start another harvester continuing where current harvester was stopped.

PR #294, opened 2015-11-26
Partial_line_waiting Removal

ruflin:
Fixes current partial line implementation which is broken by removing partial_line_waiting completely. See https://github.com/elastic/filebeat/issues/296

### DIFF
--- a/harvester/log.go
+++ b/harvester/log.go
@@ -81,20 +81,18 @@ func (h *Harvester) Harvest() {
 	//      timedReader provides timestamp some bytes have actually been read from file
 	lastReadTime := time.Now()
 
-	// remember size of last partial line being sent. Do not publish partial line, if
-	// no new bytes have been processed
-	lastPartialLen := 0
-
 	for {
-		text, bytesRead, isPartial, err := readLine(reader, &timedIn.lastReadTime, h.Config.PartialLineWaitingDuration)
+		// Partial lines return error and are only read on completion
+		text, bytesRead, err := readLine(reader, &timedIn.lastReadTime)
 
 		if err != nil {
 
 			// In case of err = io.EOF returns nil
 			err = h.handleReadlineError(lastReadTime, err)
 
+			// Return in case of error which leads to stopping harvester and closing file
 			if err != nil {
-				logp.Err("File reading error. Stopping harvester. Error: %s", err)
+				logp.Info("Read line error: %s", err)
 				return
 			}
 
@@ -105,18 +103,6 @@ func (h *Harvester) Harvest() {
 
 		// Reset Backoff
 		h.backoff = h.Config.BackoffDuration
-
-		if isPartial {
-			if bytesRead <= lastPartialLen {
-				// drop partial line event, as no new bytes have been consumed from
-				// input stream
-				continue
-			}
-
-			lastPartialLen = bytesRead
-		} else {
-			lastPartialLen = 0
-		}
 
 		// Sends text to spooler
 		event := &input.FileEvent{
@@ -129,11 +115,9 @@ func (h *Harvester) Harvest() {
 			Text:         &text,
 			Fields:       &h.Config.Fields,
 			Fileinfo:     &info,
-			IsPartial:    isPartial,
 		}
-		if !isPartial {
-			h.Offset += int64(bytesRead) // Update offset if complete line has been processed
-		}
+
+		h.Offset += int64(bytesRead) // Update offset if complete line has been processed
 
 		event.SetFieldsUnderRoot(h.Config.FieldsUnderRoot)
 		h.SpoolerChan <- event // ship the new event downstream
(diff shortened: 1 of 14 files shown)


harvester/reader.go
```diff
@@ -117,6 +117,7 @@ func (l *lineReader) next() ([]byte, int, error) {
 	bytes, err := l.outBuffer.Collect(l.outBuffer.Len())
 	l.outBuffer.Reset()
 	if err != nil {
+		// This should never happen as otherwise we have a borken state
```
urso: s/borken/broken/

harvester/log.go
```diff
@@ -310,31 +310,53 @@ func readLine(
...
+				if size == 0 && err == io.EOF {
+					logp.Debug("harvester", "errrrrr happened: %v", err)
+					return "", 0, false, err
+				}
+
+				// test for no file updates longer than partialLineWaiting
+				if time.Since(*lastReadTime) >= partialLineWaiting {
+					// return all bytes read for current line to be processed.
+					// Line might grow with further read attempts
+					line, size, err = reader.partial()
```
urso: why read partial again? we already have the result on line 320

harvester/log.go
```diff
@@ -310,31 +310,53 @@ func readLine(
...
+					logp.Debug("harvester", "Returning partial line as partial line weiating time was exceeded. Line size: %v", size)
+					return readlineString(line, size, true)
+				} else {
+
+					// Reset line to make sure nothing is sent yet.
+					size = 0
+				}
+
+				logp.Debug("harvester", "Returning partial line as partial line weiating time was exceeded. Line size: %v", size)
```
urso: do we ever return EOF? If not, this potentially disabled the backoff support

harvester/log.go
```diff
@@ -310,31 +310,53 @@ func readLine(
...
-		line, sz, err := reader.next()
+		line, size, err := reader.next()
 		if err != nil {
+
+			// EOF reached, can mean partial line or EOF
 			if err == io.EOF {
-				return "", 0, false, err
+				// return all bytes read for current line to be processed.
+				// Line might grow with further read attempts
+				_, size, lineErr := reader.partial()
```
urso: in order to get old 'partial' size, reader.partial() should be called before reader.next().

harvester/log.go
```diff
@@ -310,31 +310,53 @@ func readLine(
...
+
+				logp.Debug("harvester", "lineErr: %v", lineErr)
+
+				if size == 0 && err == io.EOF {
+					logp.Debug("harvester", "errrrrr happened: %v", err)
+					return "", 0, false, err
```
urso: so we return EOF only if we have no partial content has been read. this disables backoff if we have at least one byte in the buffer.

harvester/log.go
ruflin: Is the size returned only the bytes read since the last time partial was called or all bytes? same question for line?

harvester/log.go
ruflin: yes, but it is only disabled until partialWaitingTime is reached, right?

harvester/log.go
ruflin: will refactor that part.

harvester/log.go
ruflin: We do on line 324. But probably some unit tests would be good to confirm that we hit all the cases.

harvester/log.go
urso: line will be fully decoded line (since last incomplete read) and sz will number of input bytes consumed (since last incomplete read)

harvester/log.go
urso: the partial line handling will return err=nil => no backoff triggered

ruflin:
After further investigations and conversations with @urso the following suggest:
* In the current version of Filebeat (1.0.0) partial_line_waiting does not work as intended
* Implementing partial_line_waiting adds lots of complexity without a clear advantage
* I suggest to remove the configuration completely. The new behaviour would be the same as the behaviour of 1.0.0
  * If an incomplete line is found, the offset stays at the beginning of the line until the line is completed
  * In case the line is never completed, it will be never sent
  * Reasoning behind this: Filebeat should not be responsible logging errors which lead to incomplete lines

harvester/reader.go
```diff
@@ -206,33 +207,3 @@ func (l *lineReader) decode(end int) (int, error) {
 	l.byteCount += start
 	return start, err
 }
-
-// partial returns current state of decoded input bytes and amount of bytes
-// processed so far. If decoder has detected an error in input stream, the error
-// will be returned.
-func (l *lineReader) partial() ([]byte, int, error) {
```
ruflin: @urso I feel a little bit sorry to remove these beautiful methods.

ruflin:
New implementation done.

ruflin:
I just merged this one first: https://github.com/elastic/filebeat/pull/235 Now needs rebase and cleanup again.

ruflin:
Rebased an merged with #235. Ready for review.

tsg:
LGTM, a review from @urso would be good.

urso:
LGTM. But I'd like to see the linereader type to be moved to harvester/encoding package

ruflin:
@urso Would it be ok for you if I do that in an other PR as I plan to do some more refactoring?

urso:
+1

PR #299, opened 2015-11-27
Stdin refactoring

ruflin:
* Better split up stdin and log input_type
* Move reader to encoding

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -66,6 +66,17 @@ type HarvesterConfig struct {
 	ForceCloseFiles    bool `yaml:"force_close_files"`
 }
 
+const (
+	LogInputType   = "log"
+	StdinInputType = "stdin"
+)
+
+// List of valid input types
+var ValidInputType = map[string]struct{}{
+	StdinInputType: {},
+	LogInputType:   {},
+}
+
 // getConfigFiles returns list of config files.
 // In case path is a file, it will be directly returned.
 // In case it is a directory, it will fetch all .yml files inside this directory
--- a/crawler/prospector.go
+++ b/crawler/prospector.go
@@ -36,7 +36,6 @@ func (p *Prospector) Init() error {
 	}
 
 	return nil
-
 }
 
 // Setup Prospector Config
@@ -77,9 +76,11 @@ func (p *Prospector) setupHarvesterConfig() error {
 	}
 
 	// Setup InputType
-	if config.InputType == "" {
+	if _, ok := cfg.ValidInputType[config.InputType]; !ok {
+		logp.Info("Invalid input type set: %v", config.InputType)
 		config.InputType = cfg.DefaultInputType
 	}
+	logp.Info("Input type set to: %v", config.InputType)
 
 	config.BackoffDuration, err = getConfigDuration(config.Backoff, cfg.DefaultBackoff, "backoff")
 	if err != nil {
@@ -129,27 +130,20 @@ func (p *Prospector) Run(spoolChan chan *input.FileEvent) {
 
 	p.running = true
 
-	// Handle any "-" (stdin) paths
-	for i, path := range p.ProspectorConfig.Paths {
-
-		logp.Debug("prospector", "Harvest path: %s", path)
-
-		if path == "-" {
-			// Offset and Initial never get used when path is "-"
-			h, err := harvester.NewHarvester(
-				p.ProspectorConfig, &p.ProspectorConfig.Harvester,
-				path, nil, spoolChan)
-			if err != nil {
-				logp.Err("Error initializing harvester: %v", err)
-				return
-			}
-
-			h.Start()
-
-			// Remove it from the file list
-			p.ProspectorConfig.Paths = append(p.ProspectorConfig.Paths[:i], p.ProspectorConfig.Paths[i+1:]...)
-		}
+	logp.Info("Starting prospector of type: %v", p.ProspectorConfig.Harvester.InputType)
+	switch p.ProspectorConfig.Harvester.InputType {
+	case cfg.StdinInputType:
+		p.stdinRun(spoolChan)
+		return
+	case cfg.LogInputType:
+		p.logRun(spoolChan)
+		return
 	}
+
+	logp.Info("Invalid prospector type: %v")
+}
+
+func (p *Prospector) logRun(spoolChan chan *input.FileEvent) {
 
 	// Seed last scan time
 	p.lastscan = time.Now()
@@ -169,7 +163,6 @@ func (p *Prospector) Run(spoolChan chan *input.FileEvent) {
 		newlastscan := time.Now()
 
 		for _, path := range p.ProspectorConfig.Paths {
-			// Scan - flag false so new files always start at beginning TODO: is this still working as expected?
 			p.scan(path, spoolChan)
 		}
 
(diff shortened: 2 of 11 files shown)


config/config.go
```diff
@@ -66,6 +66,12 @@ type HarvesterConfig struct {
 	ForceCloseFiles    bool `yaml:"force_close_files"`
 }
 
+// List of valid input types
+var ValidInputType = map[string]struct{}{
+	"log":   {},
+	"stdin": {},
```
urso: maybe store some kind of constructor function in map or some 'id' to be returned? Or build map in 'Run' method. Just don't like comparisons based on strings like done on line 135.

crawler/prospector.go
```diff
@@ -77,9 +76,11 @@ func (p *Prospector) setupHarvesterConfig() error {
 	}
 
 	// Setup InputType
-	if config.InputType == "" {
+	if _, ok := cfg.ValidInputType[config.InputType]; !ok {
+		logp.Info("Invalid input type set: %v", config.InputType)
 		config.InputType = cfg.DefaultInputType
```
urso: config file invalid. We really want to fallback to some 'default' instead of failing?

ruflin: I suggest for this one yes for "BC" compatibility and also like this people don't need to specify it.

crawler/prospector.go
```diff
@@ -193,6 +186,37 @@ func (p *Prospector) Run(spoolChan chan *input.FileEvent) {
...
+	h.Start()
+
+	for {
+		if !p.running {
+			break
+		}
+		// Wait time during endless loop
+		oneSecond, _ := time.ParseDuration("1s")
+		time.Sleep(oneSecond)
+	}
```
urso: why is this loop required?

ruflin: To keep it running, as otherwise the prospector shuts down as the harvester is running inside a go routine. The nicer way would be to wait for the completion of the harvester with a channel. That comes in a next stage of refactoring.

config/config.go
ruflin: I see the problem not sure what you mean with "constructor function in map". can you make an example?

crawler/prospector.go
urso: Maybe signal back if prospectors go down. e.g. if only stdin is configured and prospector goes down we can kill filebeat. This way filebeat can be used to push a complete file once.

crawler/registrar.go
```diff
@@ -117,7 +117,7 @@ func (r *Registrar) processEvents(events []*FileEvent) {
 		}
 
 		// skip stdin
-		if *event.Source == "-" {
+		if event.InputType == "stdin" {
```
urso: some constants would be nice. Errors in names will never be caught by compiler.

crawler/prospector.go
ruflin: Agree, but also this signaling I would do in a second stage of refactoring to keep the changes small and digestable.

harvester/encoding/reader.go
```diff
@@ -35,7 +35,7 @@ type lineReader struct {
 
 const maxConsecutiveEmptyReads = 100
 
-func newTimedReader(reader io.Reader) *timedReader {
+func NewTimedReader(reader io.Reader) *timedReader {
```
urso: TimedReader is not required by encoding. It's specific to the harvester to figure times the file could be last read from (used by harvester main loop)

ruflin: Good point, will move it out again.

harvester/encoding/reader.go
```diff
@@ -14,13 +14,13 @@ import (
...
 	reader       io.Reader
-	lastReadTime time.Time // last time we read some data from input stream
+	LastReadTime time.Time // last time we read some data from input stream
 }
 
 // lineReader reads lines from underlying reader, decoding the input stream
 // using the configured codec. The reader keeps track of bytes consumed
 // from raw input stream for every decoded line.
-type lineReader struct {
+type LineReader struct {
```
urso: maybe introduce an interface 'LineReader' somewhere in harvester? Then we can implement MultilineReader based on the interface + MultilineReader will basically wrap 'LineReader' for collecting multiple lines. This hides reading implementation details from main harvester.

harvester/encoding/reader.go
ruflin: Moving it out would mean, timedReader has to be visible outside the package. Is this ok for you?

harvester/log.go
```diff
@@ -83,7 +83,7 @@ func (h *Harvester) Harvest() {
 
 	for {
 		// Partial lines return error and are only read on completion
-		text, bytesRead, err := readLine(reader, &timedIn.lastReadTime)
+		text, bytesRead, err := readLine(reader, &timedIn.LastReadTime)
```
urso: crazy. This works? timedReader was not exported from encoding package. So you have full access to the type when using 'type inference', but can not explicitly use the type its name?

crawler/prospector.go
urso: we need this kind of BC for 1.1 ?

crawler/prospector.go
urso: :+1:

harvester/encoding/reader.go
urso: well, you can keep it in harvester module. Not sure if it's required somewhere else. Why has it to be visible?

harvester/log.go
ruflin: It seems to work :-)

config/config.go
urso: constructor function not really required, but e.g. run function:

type runner func(*Prospector, chan *input.FileEvent)

var InputTypes = map[string]runner {
    "log": Prospector.logRun,
    "stdin": Prospector.stdinRun,
}

Methods for type T implicitly implement a function with type 'T' as first argument.

harvester/encoding/reader.go
ruflin: You would also move timedReader out of encoding?

harvester/encoding/reader.go
ruflin: Could make sense with the implementation of MultilineReader. At the moment I'm not even sure if we are going to need and additional interface so I would keep it for the moment.

harvester/encoding/reader.go
urso: yes. It's not required by encoding package.

crawler/prospector.go
ruflin: My concern is that it literally breaks all configs which have type inside or have it commented out. I will add a note to remove this in the next major version.

crawler/registrar.go
ruflin: +1

crawler/prospector.go
urso: :+1:

harvester/encoding/reader.go
ruflin: Move out.

config/config.go
ruflin: This is a very nice solution as soon as stdin and log are properly separated, means both are based on Prospector Interface. This is not the case (yet). I suggest to go at the moment with "constants" and move it to this solution as soon as the abstraction is done.

crawler/registrar.go
ruflin: Implemented through constants

ruflin:
@urso Feedback implemented. I suggest to squash again before merging.

crawler/prospector.go
```diff
@@ -129,27 +130,20 @@ func (p *Prospector) Run(spoolChan chan *input.FileEvent) {
...
-			h, err := harvester.NewHarvester(
-				p.ProspectorConfig, &p.ProspectorConfig.Harvester,
-				path, nil, spoolChan)
-			if err != nil {
-				logp.Err("Error initializing harvester: %v", err)
-				return
-			}
+	logp.Info("Starting prospector of type: %v", p.ProspectorConfig.Harvester.InputType)
+	switch p.ProspectorConfig.Harvester.InputType {
+	case "stdin":
```
urso: Use your constants?

ruflin: done

urso:
LGTM

ruflin:
@urso Squashed

PR #302, opened 2015-11-30
Improve force_close_file config based on discussion in #93

ruflin:
Add additional tests to check for improvement.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -147,11 +147,13 @@ lines. The `backoff` value will be multiplied each time with the `backoff_factor
 
 ===== force_close_files
 
-By default, Filebeat keeps the files that it’s reading open until the timespan specified by `ignore_older` has elapsed. This behaviour can cause issues when a file is removed. Because the file isn't fully removed until Filebeat closes the file, no new file with the same name can be created during this time.
+By default, Filebeat keeps the files that it’s reading open until the timespan specified by `ignore_older` has elapsed.
+This behaviour can cause issues when a file is removed. On Windows, the file cannot be fully removed until Filebeat closes
+the file. In addition no new file with the same name can be created during this time.
 
-You can force Filebeat to close the file as soon as the file name changes by setting the `force_close_windows_files` option to true. This option is supported on Windows only. The default is false.
-
-Turning on this option can lead to loss of data on rotated files. After file rotation, the beginning of the new file might be skipped because the reading starts at the end of the file. We recommend that you leave this option set to false, and instead specify a lower value for the `ignore_older` option to release files faster.
+You can force Filebeat to close the file as soon as the file name changes by setting the
+`force_close_files` option to true. The default is false. Turning on this option can lead to loss of data on
+rotated files in case not all lines were read from the rotated file.
 
 ===== spool_size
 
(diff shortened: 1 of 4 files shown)


docs/configuration.asciidoc
```diff
@@ -147,11 +147,16 @@ lines. The `backoff` value will be multiplied each time with the `backoff_factor
...
+the file, no new file with the same name can be created during this time.
 
-You can force Filebeat to close the file as soon as the file name changes by setting the `force_close_windows_files` option to true. This option is supported on Windows only. The default is false.
+You can force Filebeat to close the file as soon as the file name changes by setting the `force_close_windows_files` option to true.
+This option is supported on Windows only. The default is false.
 
-Turning on this option can lead to loss of data on rotated files. After file rotation, the beginning of the new file might be skipped because the reading starts at the end of the file. We recommend that you leave this option set to false, and instead specify a lower value for the `ignore_older` option to release files faster.
+Turning on this option can lead to loss of data on rotated files. After file rotation, the beginning of the
+new file might be skipped because the reading starts at the end of the file. We recommend that you leave this
+option set to false, and instead specify a lower value for the `ignore_older` option to release files faster.
```
tsg: Since we know that a low ignore_older causes issues as well, we might want to remove this recommendation for now?

ruflin: Good point, will update it.

ruflin: Updated and some more improvements made to the docs so they are more accurate.

harvester/log.go
```diff
@@ -271,10 +271,12 @@ func (h *Harvester) handleReadlineError(lastTimeRead time.Time, err error) error
...
 	}
 
-	// On windows, check if the file name exists (see #93)
 	if h.Config.ForceCloseFiles {
+		// Check if the file name exists (see #93)
 		_, statErr := os.Stat(h.file.Name())
-		if statErr != nil {
+
+		// Error means file does not exist. If no error, check if same file. If not close as rotated.
+		if statErr != nil || !input.IsSameFile(h.file.Name(), info) {
```
tsg: Hmm, now sure I get this. IsSameFile seems to use `os.Stat()` which will for sure return an error and hence always be false.

urso: it's using or. only if statErr==nil we will check IsSameFile

docs/configuration.asciidoc
```diff
@@ -147,11 +147,13 @@ lines. The `backoff` value will be multiplied each time with the `backoff_factor
...
+By default, Filebeat keeps the files that it’s reading open until the timespan specified by `ignore_older` has elapsed.
+This behaviour can cause issues when a file is removed. On Windows, the file cannot be fully removed until Filebeat closes
+the file. In addition no new file with the same name can be created during this time.
 
-You can force Filebeat to close the file as soon as the file name changes by setting the `force_close_windows_files` option to true. This option is supported on Windows only. The default is false.
-
-Turning on this option can lead to loss of data on rotated files. After file rotation, the beginning of the new file might be skipped because the reading starts at the end of the file. We recommend that you leave this option set to false, and instead specify a lower value for the `ignore_older` option to release files faster.
+You can force Filebeat to close the file as soon as the file name changes by setting the
+`force_close_windows_files` option to true. The default is false. Turning on this option can lead to loss of data on
```
urso: force_close_windows_files?

ruflin: fixed

PR #304, opened 2015-12-01
Add win installsteps

dedemorton:
Added windows steps. If we like this approach, we should do this for all Beats (move the installation steps into the GS). 

I think it's good to have all the core steps to get up and running in one document, even though it makes the document a tad longer.

### DIFF
--- a/docs/getting-started.asciidoc
+++ b/docs/getting-started.asciidoc
@@ -12,10 +12,15 @@ After you have installed these products, you can start <<filebeat-installation>>
 [[filebeat-installation]]
 === Installing Filebeat
 
-To download and install Filebeat, run the following commands. Use the commands that
-work with your system (deb for Debian/Ubuntu, rpm for Redhat/Centos/Fedora, mac for OS X).
+To download and install Filebeat, use the commands that work with your system
+(<<deb, deb>> for Debian/Ubuntu, <<rpm, rpm>> for Redhat/Centos/Fedora, <<mac,
+mac>> for OS X, and <<win, win>> for Windows).
 
-deb:
+NOTE: We also provide 32-bit images that you can get from our
+https://www.elastic.co/downloads/beats/filebeat[download page].
+
+[[deb]]
+==== deb:
 
 ["source","sh",subs="attributes,callouts"]
 ------------------------------------------------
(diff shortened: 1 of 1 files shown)


monicasarbu:
@dedemorton I really like the result. Maybe we should do the same thing for the other Beats as well if you have the time. Thanks!

ruflin:
:+1: on the result. Makes all of the installations look like equal citizens. Thanks.

docs/getting-started.asciidoc
```diff
@@ -160,29 +188,41 @@ Filebeat, you need to set up Filebeat to use Logstash. For detailed steps, see
...
+==== win:
+
+[source,shell]
+----------------------------------------------------------------------
+PS > Start-Service filebeat
+----------------------------------------------------------------------
+
+By default, Windows log files are stored in `C:\ProgramData\filebeat\Logs`.
+
+NOTE: On Windows, the statistics about system load and swap usage are currently
```
andrewkroh: I think this note is specific to Topbeat.

docs/getting-started.asciidoc
```diff
@@ -42,10 +46,25 @@ curl -L -O https://download.elastic.co/beats/filebeat/filebeat-{version}-darwin.
...
-NOTE: We also provide 32-bit images that you can get from our
-https://www.elastic.co/downloads/beats/filebeat[download page].
+[[win]]
+==== win:
+
+. Download the Filebeat Windows zip file from the
+https://www.elastic.co/downloads/beats/filebeat[downloads page].
+
+. Extract the contents of the zip file to your computer, for example, to `C:\Program Files`.
+
```
andrewkroh: The commands work as is, but I think that the document should advise renaming the extracted folder to `Filebeat` such that the installation path is `C:\Program Files\Filebeat`. The reason is that by default if you just run the executable without specifying a config file location like `.\filebeat.exe -configtest` it will automatically find the configuration file located at `C:\Program Files\Filebeat\filebeat.yml`.

docs/getting-started.asciidoc
```diff
@@ -42,10 +46,25 @@ curl -L -O https://download.elastic.co/beats/filebeat/filebeat-{version}-darwin.
...
+. Open a PowerShell prompt as an Administrator (right-click the PowerShell icon and select *Run As Administrator*), and navigate to where you extracted the
+zip file.
```
andrewkroh: I think it would be nice to show the `cd` command.

`PS > cd C:\Program Files\Filebeat`

ruflin:
For reference: Changes were made here: https://github.com/elastic/filebeat/pull/305

PR #346, opened 2016-03-14
spelling mistake - independent

azhurbilo:


ruflin:
@azhurbilo Thank you for the contribution. This repository is not active anymore as all development to filebeat happens in the beats repository. The typo should be fixed there already (https://github.com/elastic/beats/blob/master/filebeat/etc/beat.yml#L124) but if you spot some more please open a PR to the beats repo.

